PR #3, opened 2023-08-29
migrate repo over

szaldivar:
* Also ran prettier in all of the repo; that is why the docs also changed, only changed the format. We are still working on a docs publishing pipeline.


Lu-Whale (approved):
LGTM

JordanOcokoljic (approved):
In all honesty we should probably mark `docs` as an ignored directory but we can worry about that later. LGTM

src/psychrometrics/p_sat_torr.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Estimates the saturation vapour pressure in [torr].
+ *
+ * Estimates the saturation vapour pressure in [torr].
```
FedericoTartarini, line 4: It looks like here there is a repetition in the code.

src/psychrometrics/t_dp.js
```diff
@@ -0,0 +1,22 @@
+import { round } from "../utilities/utilities";
+
+/**
+ * Calculates the dew point temperature.
+ * @param {number} bulbTemperature - dry bulb air temperature, [°C]
```
FedericoTartarini, line 5: we will need to use a consistent naming for the input variables. bulbTemperature is not the best variable name. We can use the same (or similar) naming convention that we are currently using in pythermalcomfort

src/utilities/utilities.js
```diff
@@ -216,6 +216,185 @@ export function body_surface_area(weight, height, formula = "dubois") {
...
 }
+
+/**
+ * Estimates the relative air speed which combines the average air speed of the
+ * space plus the relative air speed caused by the body movement. Vag is assumed
+ * to be 0 for metabolic rates equal and lower than 1 met and otherwise equal to
+ * Vag = 0.3 (M - 1) (m/s)
+ *
+ * @template {(number | number[])} T
+ * @param {T} v - air spped measured by the sensor, [m/s]
```
FedericoTartarini, line 230: There are typos in this line and T is not the right variable name for airSpeed, see my previous comment and let's discuss this during our next meeting

src/utilities/utilities.js
```diff
@@ -216,6 +216,185 @@ export function body_surface_area(weight, height, formula = "dubois") {
...
+  }
+  return _v_relative_single(v, met);
+}
+
+/**
+ * @param {number} v
+ * @param {number} met
+ * @returns {number}
+ */
+function _v_relative_single(v, met) {
```
FedericoTartarini, line 247: The name of this function is not very informative

src/utilities/utilities.js
```diff
@@ -216,6 +216,185 @@ export function body_surface_area(weight, height, formula = "dubois") {
...
+/**
+ * Estimates the dynamic clothing insulation of a moving occupant. The activity as
+ * well as the air speed modify the insulation characteristics of the clothing and the
+ * adjacent air layer. Consequently, the ISO 7730 states that the clothing insulation
+ * shall be corrected [2]_. The ASHRAE 55 Standard corrects for the effect
+ * of the body movement for met equal or higher than 1.2 met using the equation
+ * clo = Icl × (0.6 + 0.4/met)
+ *
+ * @template {(number | number[])} T
+ * @param {T} clo - clothing insulation, [clo]
```
FedericoTartarini, line 260: Here again there is the same issue where it says @param {T} but we are referring to clothing insulation

FedericoTartarini: See also the lines below that have issues

src/utilities/utilities.js
```diff
@@ -216,6 +216,185 @@ export function body_surface_area(weight, height, formula = "dubois") {
...
+    return clo;
+  return _clo_dynamic_single(clo, met);
+}
+
+/**
+ * @param {number} clo
+ * @param {number} met
+ * @returns {number}
+ */
+function _clo_dynamic_single(clo, met) {
```
FedericoTartarini, line 294: we also need to find a better name for this function. I am not sure why you are appending _single at the end of the function name

FedericoTartarini (commented):
Great work, I left a few comments for you to address. Please also review the other sections of the code since some of the issues I mentioned in the code may be present in other sections of the code that I have not commented.

src/utilities/utilities.js
JordanOcokoljic: `T` here is not the variable name, it is the type of the provided parameter. `T` is defined by the `@template` tag on line 229.

FedericoTartarini: I am very sorry about that. Would it be possible to give this template a more meaningful and informative name?

JordanOcokoljic: While it is possible to give the template any name you'd please, `T` is the accepted industry standard. Historically this is done to ensure that a compile time name is not confused with, for example, a parameter or variable name.

src/psychrometrics/p_sat_torr.js
JordanOcokoljic: Looks like duplication due to a messy merge - we'll remove that!

src/utilities/utilities.js
JordanOcokoljic: `_single` has been appended because this function performs the calculation for `clo_dynamic` for a single value. This function isn't made public, and is only used internally within the `export`ed `clo_dynamic` function as defined on line 265, which can accept either an array of values or a single value, and will return an array or a single number based on it's input.

The reason that this is a little thorny is that in the world of Python the primary pattern is the iterator, so most functions accept either lists or single values and any differing treatment of them internally is opaque to the caller. In the world of Javascript, this is not the case, and the ways of handling this are:

1. Have two separate `export`ed functions, one for handling the array case, and one for handling the singular case. This is by far the most common.
2. Have one `export`ed function, which only accepts arrays, and in the documentation specify that if the caller wants to provide a single value, they should provide an array with only that element inside it. This one is uncommon, but still occasionally used.
3. Have one `export`ed function, which accepts either arrays or single values, and handle the singular or array case by using `Array.isArray` going down a separate path in the function. This is generally not seen in large libraries, because it is harder to reason about exactly what a function will do just by scanning the code quickly.

However, to keep parity with `pythermalcomfort` we have opted for this last option. In order to keep the code maintainable and avoid duplication we have extracted the core calculation behind `clo_dynamic` into this un-exported function (which we have opted to name `clo_dynamic_single` because it makes it clearer - to us at least - that it operates on single values). In the `export`ed we either choose to just call our single function on the provided parameter (in the case that it is singular) and otherwise we call it on each element in the provided array (in the case that an array has been provided).

Hope this clarifies our thought process around this.

src/utilities/utilities.js
JordanOcokoljic: See my comment on `clo_dynamic_single` for the rationale around this decision.

src/utilities/utilities.js
FedericoTartarini: Great, thank you so much for the clear explanation, very happy with your solution. Maybe let's just add a line of comments under the function called single to inform future developers about this solution.

PR #4, opened 2023-08-31
add check standard compliance array version and t_o array version

szaldivar:


tests/psychrometrics/t_o.test.js
```diff
@@ -95,7 +95,7 @@ describe("t_o", () => {
   );
 
   it("throws an error if the airSpeed is negative", () => {
-    expect(() => t_o(0, 0, -1, "ISO")).toThrow("airSpeed cannot be negative");
+    expect(() => t_o(0, 0, -1, "ISO")).toThrow("v cannot be negative");
```
JordanOcokoljic, line 98: I think we should have some tests here for the array version of the function - just so we have confidence moving forward that we haven't broken things accidentally (see [Test First](https://www.notion.so/Test-First-6fe7edeb7c424182ba49dc973549bbd0?pvs=4))

src/utilities/utilities.js
```diff
@@ -47,6 +63,90 @@ export function check_standard_compliance(standard, kwargs) {
...
+ * Check that the values as an array comply with the standard provided and returns arrays where
+ * the values that do not comply are NaN
+ * @see {@link check_standard_compliance} for scalar variant that returns warnings
+ *
+ * @param {Standard | "fan_heatwaves"} standard - standard to check compliance with
+ * @param {ComplianceKwargsArray & {airspeed_control?: boolean}} kwargs - values to check compliance against
+ *
+ * @returns {CheckStandardComplianceResult} filtered arrays based on compliance limits
+ */
+export function check_standard_compliance_array(standard, kwargs) {
```
JordanOcokoljic, line 86: This looks _really_ long, and I know I wouldn't be confident changing it unless I had some tests to assure me that I hadn't broken anything along the way.

src/psychrometrics/t_o.js
```diff
@@ -1,36 +1,78 @@
...
-      );
+      return _t_o_calculation_ashrae(tdb, tr, v);
     default:
       throw new Error("standard must be one of ISO or ASHRAE");
   }
 }
+
+/**
+ * Calculates operative temperature in accordance with ISO 7726:1998.
+ * Accepts array arguments. @see {@link t_o} for scalar arguments
```
JordanOcokoljic, line 28: Nothing to change - just wanted to say that I love this comment and think it should be the way moving forward we clear up the differences between the scalar and vector functions.

tests/psychrometrics/t_o.test.js
szaldivar: yeah, good point, I will add them

src/utilities/utilities.js
szaldivar: yep, will add them

JordanOcokoljic (approved):
Looks good to me!

PR #8, opened 2023-09-02
finish t_mrt

Ilite-Kiriya:


szaldivar:
hey, if you want us to review this PR please request our review in the PR so we get a notification

src/psychrometrics/t_mrt.js
```diff
@@ -0,0 +1,162 @@
+import { valid_range } from "../utilities/utilities";
+
+const g = 9.81;
+const c_to_k = 273.15;
+const cp_air = 1004;
+
+/**
+ * Converts globe temperature reading into mean radiant temperature in accordance with either the Mixed Convection developed by Teitelbaum E. et al. (2022) or the ISO 7726:1998 Standard [5]_.
+ * @param {Number | Number[]} tg - globe temperature, [°C]
```
szaldivar, line 9: `number` rather than `Number`

src/psychrometrics/t_mrt.js
```diff
@@ -0,0 +1,162 @@
...
+ * @param {Number | Number[]} tdb - air temperature, [°C]
+ * @param {Number | Number[]} v - air speed, [m/s]
+ * @param {Number | Number[]} d - diameter of the globe, [m] default 0.15 m
+ * @param {Number | Number[]} emissivity - emissivity of the globe temperature sensor, default 0.95
+* @param {String} standard - {"Mixed Convection", "ISO"} either choose between the Mixed Convection and ISO formulations.
```
szaldivar, line 14: change `String` to `"Mixed Convection" | "ISO"` so our users get intellisense/autocomplete with the options

src/psychrometrics/t_mrt.js
```diff
@@ -0,0 +1,162 @@
...
+  standard = "Mixed Convection",
+) {
+  standard = standard.toLowerCase();
+  tdb = Array.isArray(tdb) ? tdb : [tdb];
+  tg = Array.isArray(tg) ? tg : [tg];
+  v = Array.isArray(v) ? v : [v];
+  d = Array.isArray(d) ? d : [d];
+  let tr = 0;
+
+  if (standard === "mixed convection") {
```
szaldivar, line 39: can you move the code inside each if to their own functions? something like `_t_mrt_mixed_convection`

src/psychrometrics/t_mrt.js
```diff
@@ -0,0 +1,162 @@
...
+export function t_mrt(
+  tg,
+  tdb,
+  v,
+  d = 0.15,
+  emissivity = 0.95,
+  standard = "Mixed Convection",
+) {
+  standard = standard.toLowerCase();
+  tdb = Array.isArray(tdb) ? tdb : [tdb];
```
szaldivar, line 33: I think it is better to create a variable for `tdbIsArray` and don't convert `tdb` into an array. Then use that variable inside the map `const _tdb = tdbIsArray ? tdb[index] : tdb`
Makes it much cleared why you are doing it than checking the length of the array to be 1. Also having that `length === 1` check means that if the user gives us input that should fail like `tg = [1], tdb = [1,2]` then the function would not fail when it probably should since that input means something went wrong in the user's end, like missing data etc.

src/psychrometrics/t_mrt.js
```diff
@@ -0,0 +1,162 @@
...
+          });
+
+    tr = tg.map((_tg, i) => {
+      const _tdb = tdb.length === 1 ? tdb[0] : tdb[i];
+      const _emissivity = !Array.isArray(emissivity)
+        ? emissivity
+        : emissivity[i];
+      const _h = h.length === 1 ? h[0] : h[i];
+      return (
+        Math.round(
```
szaldivar, line 148: we have a round function for doing this, check utilities.js

tests/psychrometrics/t_mrt.test.js
```diff
@@ -0,0 +1,34 @@
...
+      tdb: 26.1,
+      v: 0.1931,
+      d: [0.1, 0.1, 0.5, 0.03],
+      emissivity: 0.95,
+      standard: "Mixed Convection",
+      expected: [24.2, 27.0, NaN, NaN],
+    },
+  ])("test t_mrt", ({ tg, tdb, v, d, emissivity, standard, expected }) => {
+    const result = t_mrt(tg, tdb, v, d, emissivity, standard);
+    for (let i = 0; i < expected.length; i++) {
```
szaldivar, line 26: we have a function for this case, see `test_utilities.js` inside the `tests` directory

szaldivar (changes requested):

src/psychrometrics/t_mrt.js
```diff
@@ -0,0 +1,162 @@
...
+ * @param {Number | Number[]} d - diameter of the globe, [m] default 0.15 m
+ * @param {Number | Number[]} emissivity - emissivity of the globe temperature sensor, default 0.95
+* @param {String} standard - {"Mixed Convection", "ISO"} either choose between the Mixed Convection and ISO formulations.
+        The Mixed Convection formulation has been proposed by Teitelbaum E. et al. (2022)
+        to better determine the free and forced convection coefficient used in the
+        calculation of the mean radiant temperature. They also showed that mean radiant
+        temperature measured with ping-pong ball-sized globe thermometers is not reliable
+        due to a stochastic convective bias [22]_. The Mixed Convection model has only
+        been validated for globe sensors with a diameter between 0.04 and 0.15 m.
+* @returns {Number | Array}
```
szaldivar, line 21: use `number[]` rather than `Array`

szaldivar (changes requested):

src/psychrometrics/t_mrt.js
```diff
@@ -0,0 +1,164 @@
...
+    return (
+      2 +
+      (0.4 * Math.pow(_re, 0.5) + 0.06 * Math.pow(_re, 2 / 3)) *
+        Math.pow(pr, 0.4)
+    );
+  }); //arr, size same as d
+
+  tr = tg.map((_tg, i) => {
+    const _tdb = tdbIsArray ? tdb[i] : tdb;
+    const _nu_forced = nu_forced.length === 1 ? nu_forced[0] : nu_forced[i];
```
szaldivar, line 89: do the same as tdb here, instead of checking the length

src/psychrometrics/t_mrt.js
```diff
@@ -0,0 +1,164 @@
...
+    const _nu_natural = nu_natural.length === 1 ? nu_natural[0] : nu_natural[i];
+    const _n = n.length === 1 ? n[0] : n[i];
+    const _d = d.length === 1 ? d[0] : d[i];
+    const _emissivity = Array.isArray(emissivity) ? emissivity[i] : emissivity;
```
szaldivar, line 93: move check outside the loop, similar to `tdbIsArray`

src/psychrometrics/t_mrt.js
```diff
@@ -0,0 +1,164 @@
...
+            _emissivity /
+            o,
+        0.25,
+      ) - 273.15
+    );
+  });
+
+  const d_valid = valid_range(d, [0.04, 0.15]);
+  const trResult = d_valid.map((_d_valid, i) =>
+    !isNaN(_d_valid) ? round(tr[i], 1) : NaN,
```
szaldivar, line 113: no need for this check, `round` can handle NaNs

src/psychrometrics/t_mrt.js
```diff
@@ -0,0 +1,164 @@
...
+function t_mrt_iso(tg, tdb, v, d, emissivity) {
+  let tr = 0;
+  const tdbIsArray = Array.isArray(tdb);
+  tg = tg.length === 1 ? [tg[0] + c_to_k] : tg.map((_tg) => _tg + c_to_k);
+  tdb = tdbIsArray ? tdb.map((_tdb) => _tdb + c_to_k) : tdb + c_to_k;
+
+  // calculate heat transfer coefficient
+  const h_n = tg.map((_tg, i) => {
+    const _tdb = tdbIsArray ? tdb[i] : tdb;
+    const _d = d.length === 1 ? d[0] : d[i];
```
szaldivar, line 128: same comment here, change all these checks for length === 1 to checking a computer variable to see if they are arrays

szaldivar (changes requested):

src/psychrometrics/t_mrt.js
Ilite-Kiriya: line 111 to line 113 refers to 
d_valid = valid_range(d, (0.04, 0.15))
        tr = np.where(~np.isnan(d_valid), tr, np.nan)
as tr result is related to d, I think we should check isNan(_d_valid) to return either tr[i] or NaN

Ilite-Kiriya:
Some new observations:
As t_mrt converts globe temperature, tg will in any case have the longest length (either tg is a number, or tg.length is the largest among all array type parameters). for calculations related to tg, it's better to use tg.map()
I converted all parameters to array type, so it will be more convenient to use .map functions, otherwise, I need to not only check if Array.isArray(var), but also need to compare the length for calculating (e.g., calculate re if the standard is ISO, ), New example code is below
const re = d.length >= v.length ? d.map((_d, i) => {
    const _v = v_len1 ? v[0] : v[i];
    return (_v * _d) / nu;//size d 
  }) : v.map((_v) => (_v * d[0]) / nu) //size v
  const re_len1 = re.length === 1

If the standard is mixed convection, as the result is based on valid range of d, tg’s length should always be the same as d’s length, or both tg and d are numbers. Therefore, for the calculation below related to only d and v, I used d.map
const re = d.map((_d, i) => {
    const _v = v_len1 ? v[0] : v[i];
    return (_v * _d) / nu;//size d 
  })

Ilite-Kiriya:
Also, here are some explanations about my code, 
As t_mrt converts globe temperature, tg will in any case have the longest length (either tg is a number, or tg.length is the largest among all array type parameters)
I converted all parameters to array type, so it will be more convenient to use .map functions, otherwise, I need to not only check if Array.isArray(var), but also need to compare the length for calculating (e.g., when calculate re, if the standard is ISO, ), New code is below
const re = d.length >= v.length ? d.map((_d, i) => {
    const _v = v_len1 ? v[0] : v[i];
    return (_v * _d) / nu;//size d 
  }) : v.map((_v) => (_v * d[0]) / nu) //size v
  const re_len1 = re.length === 1

If the standard is mixed convection, as the result is based on valid range of d, tg’s length should always be the same as d’s length, or both tg and d are numbers. Therefore, for the calculation below related to only d and v, I used d.map
const re = d.map((_d, i) => {
    const _v = v_len1 ? v[0] : v[i];
    return (_v * _d) / nu;//size d 
  })

src/psychrometrics/t_mrt.js
```diff
@@ -0,0 +1,139 @@
...
+  v,
+  d = 0.15,
+  emissivity = 0.95,
+  standard = "Mixed Convection",
+) {
+  standard = standard.toLowerCase();
+  let tr = 0;
+  if (standard === "mixed convection") {
+    tr = get_tr_mixed_convection(tg, tdb, v, d, emissivity);
+    const d_valid = valid_range([d], [0.04, 0.15]);
```
szaldivar, line 43:
```suggestion
    const d_valid = d >= 0.04 && d <= 0.15 ? d : NaN;
```

src/psychrometrics/t_mrt.js
```diff
@@ -0,0 +1,139 @@
...
+    return isNaN(d_valid[0]) ? NaN : round(tr, 1);
```
szaldivar, line 44:
```suggestion
    return isNaN(d_valid) ? NaN : round(tr, 1);
```

src/psychrometrics/t_mrt.js
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  if (standard === "iso") {
+    return get_tr_iso(tg, tdb, v, d, emissivity);
+  }
```
szaldivar, line 49: throw here? no standard found

src/psychrometrics/t_mrt.js
```diff
@@ -0,0 +1,139 @@
...
+    );
+    return trResult;
+  }
+
+  if (standard === "iso") {
+    tr_array = tg.map((_tg, i) => {
+      return get_tr_iso(_tg, tdb[i], v[i], d[i], emissivity[i]);
+    });
+    return tr_array;
+  }
```
szaldivar, line 91: throw here, no standard found

szaldivar (changes requested):

src/psychrometrics/t_mrt.js
```diff
@@ -0,0 +1,139 @@
...
+const mu = 0.0000181; // Pa s
+const k_air = 0.02662; // W/m-K
+const beta = 0.0034; // 1/K
+const nu = 0.0000148; // m2/s
+const alpha = 0.00002591; // m2/s
+const pr = (cp_air * mu) / k_air; // Prandtl constants, number
+const o = 0.0000000567;
+
+/**
+ * Converts globe temperature reading into mean radiant temperature in accordance with either the Mixed Convection developed by Teitelbaum E. et al. (2022) or the ISO 7726:1998 Standard [5]_.
```
szaldivar, line 15: also please add the link to the corresponding function here, see `t_o` and `t_o_array`
make sure the `@see` is in a new line, otherwise our docs seem to not work

szaldivar (changes requested):

src/psychrometrics/t_mrt.js
```diff
@@ -0,0 +1,147 @@
...
+    return get_tr_iso(tg, tdb, v, d, emissivity);
+  }
+  throw new Error(
+    "No standard found, please choose standard from ISO and Mixed Convectioin",
+  );
+}
+
+/**
+ * Converts globe temperature reading into mean radiant temperature in accordance with either the Mixed Convection developed by Teitelbaum E. et al. (2022) or the ISO 7726:1998 Standard [5]_.
+ * Accepts array arguments. @see {@link t_mrt} for scalar arguments
```
szaldivar, line 58: can you change the `@see {@link ...` to the beginning of the line. I changed also this code in t_o in main because the docs do not generate correctly if `@see` is not at the beginning

szaldivar (changes requested):

src/psychrometrics/t_mrt.js
szaldivar: ah I meant to the beginning of the line, [see here for an example](https://github.com/FedericoTartarini/jsthermalcomfort/blob/main/src/psychrometrics/t_o.js#L28)

szaldivar (changes requested):

ChristinChen233 (approved):

ChristinChen233:
add t_mrt to main

PR #9, opened 2023-09-02
Remove JSDoc and better-docs in favour of `documentation`

JordanOcokoljic:
I'm removing JSDoc (as a generator) because it's really fallen behind the "standard" for type definitions in the ecosystem. [`documentation`](https://documentation.js.org/) seems to be a little more well maintained.

I've also left the `docs` folder in there because it makes using Github Pages a lot easier, I'm still iffy about that because there's currently not an automated way to ensure that people have updated the documentation for their PR, and it could make merge conflicts utter hell if there are ever any in there - but that is a problem to worry about when we actually run into it.

An alternative would be having a second repo and setting up an action to commit and push to that repo on successful merge, which would avoid having all that state in this repository - again, something for another day.


szaldivar:
I think we can set up a GitHub action that runs after each merge to main to run the docs generation and automatically commit it to main

szaldivar (approved):

PR #16, opened 2023-09-04
Yuchen/set_tmp,two_nodes

YuchenMu:
set_tmp and two_nodes done


szaldivar:
need to resolve conflicts

szaldivar:
tests/prettier not passing

szaldivar:
Please include only one model per PR; in this case, raise another PR with only the function that has all its dependencies ready

PR #21, opened 2023-09-05
Christin/wc

Ilite-Kiriya:


src/models/wc.js
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Calculates the Wind Chill Index (WCI) in accordance with the ASHRAE 2017 Handbook Fundamentals - Chapter 9 [18]_.
+ * @param {number} tdb - dry bulb air temperature,[°C]
+ * @param {number} v - wind speed 10m above ground level, [m/s]
+ * @param {boolean} ifRound - default True, if True rounds output value, if False it does not round it
+ * @returns {object} {"wci": wci} - wind chill index, [W/m2)]
+ */
+
+export function wc(tdb, v, ifRound = true) {
```
szaldivar, line 11: can you take kwargs and call the option round? to be consistent with the rest of the functions.
something like:
```javascript
function wc(tdb, v, {round} = {round: true}) {
}
```

szaldivar (changes requested):

szaldivar (changes requested):

src/models/wc.js
```diff
@@ -0,0 +1,19 @@
+import { round } from "../utilities/utilities.js";
+
+/**
+ * Calculates the Wind Chill Index (WCI) in accordance with the ASHRAE 2017 Handbook Fundamentals - Chapter 9 [18]_.
+ * @param {number} tdb - dry bulb air temperature,[°C]
+ * @param {number} v - wind speed 10m above ground level, [m/s]
+ * @param {object} ifRound - default True, if True rounds output value, if False it does not round it
```
szaldivar, line 7: need to update this type

szaldivar: also [see here](https://github.com/FedericoTartarini/jsthermalcomfort/blob/main/src/models/heat_index.js#L8-L9) for an example of how to annotate these types

szaldivar (changes requested):

src/models/wc.js
```diff
@@ -0,0 +1,20 @@
+import { round } from "../utilities/utilities.js";
+
+/**
+ * Calculates the Wind Chill Index (WCI) in accordance with the ASHRAE 2017 Handbook Fundamentals - Chapter 9 [18]_.
+ * @param {number} tdb - dry bulb air temperature,[°C]
+ * @param {number} v - wind speed 10m above ground level, [m/s]
+ * @param {Object} [kwargs] (Optional) Other parameters.
+ * @param {boolean} [kwargs.round=true] - If True rounds output value, if False it does not round it.
+ * @returns {object} {"wci": wci} - wind chill index, [W/m2)]
```
szaldivar, line 9: Does this work? Do you get intellisense on the result when you use the function?

szaldivar (approved):

PR #23, opened 2023-09-06
"discomfort_index" work package finished

YuchenMu:
API in both float and array versions of "discomfort_index" finished.

### DIFF
--- a/src/models/discomfort_index.js
+++ b/src/models/discomfort_index.js
@@ -1,7 +1,8 @@
 import { round } from "../utilities/utilities";
 
 /**
- * @typedef {Object} DiscomfortIndexReturnType
+ * @typedef {Object} DiscomfortIndexReturnType - a result set containing the discomfort index and the classification of the
+ * thermal comfort conditions
  * @property {number} di – Discomfort Index(DI)
  * @property {string} discomfort_condition Classification of the thermal comfort conditions according to the discomfort index
  */
@@ -30,7 +31,8 @@ export function discomfort_index(tdb, rh) {
 }
 
 /**
- * @typedef {Object} DiscomfortIndexArrayReturnType
+ * @typedef {Object} DiscomfortIndexArrayReturnType - a result set containing an array of discomfort index and an array
+ * of classification of the thermal comfort conditions
  * @property {number[]} di – Discomfort Index(DI) Array
  * @property {string[]} discomfort_condition Classification of the thermal comfort conditions in array
  */
--- a/src/models/discomfort_index.js
+++ b/src/models/discomfort_index.js
@@ -70,22 +70,19 @@ export function discomfort_index_array(tdb, rh) {
  * @returns {string} Discomfort condition
  */
 function check_categories(di) {
-  const diCategories = {
-    21: "No discomfort",
-    24: "Less than 50% feels discomfort",
-    27: "More than 50% feels discomfort",
-    29: "Most of the population feels discomfort",
-    32: "Everyone feels severe stress",
-    99: "State of medical emergency",
-  };
+  const threshold = [21, 24, 27, 29, 32, 99];
+  const condition = [
+    "No discomfort",
+    "Less than 50% feels discomfort",
+    "More than 50% feels discomfort",
+    "Most of the population feels discomfort",
+    "Everyone feels severe stress",
+    "State of medical emergency",
+  ];
 
-  let condition = "default";
-  for (const key in diCategories) {
-    if (di <= key) {
-      condition = diCategories[key];
-      break;
+  for (let index = 0; index < threshold.length; index++) {
+    if (di <= threshold[index]) {
+      return condition[index];
     }
   }
-
-  return condition;
 }
--- a/src/models/discomfort_index.js
+++ b/src/models/discomfort_index.js
@@ -7,45 +7,42 @@ import { round } from "../utilities/utilities";
  */
 
 /**
- * Calculates the Discomfort Index (DI). The index is essentially an effective temperature based on air temperature and humidity. 
+ * Calculates the Discomfort Index (DI). The index is essentially an effective temperature based on air temperature and humidity.
  * The discomfort index is usuallly divided in 6 dicomfort categories and it only applies to warm environments.
  * @see {@link discomfort_index_array} for a version that supports arrays
  *
  * @param {number} tdb - air temperature [C]
  * @param {number} rh - relative humidity [%]
- * 
+ *
  * @returns {discomfortIndexReturnType} object with results of DI
  * @example
- * const DI = discomfort_index(25, 50); // returns { di: 22.1, discomfort_condition: 'Less than 50% feels discomfort' } 
+ * const DI = discomfort_index(25, 50); // returns { di: 22.1, discomfort_condition: 'Less than 50% feels discomfort' }
  */
 export function discomfort_index(tdb, rh) {
+  const di = tdb - 0.55 * (1 - 0.01 * rh) * (tdb - 14.5);
 
-    const di = tdb - 0.55 * (1 - 0.01 * rh) * (tdb - 14.5);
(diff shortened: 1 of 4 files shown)


src/models/discomfort_index.js
```diff
@@ -0,0 +1,98 @@
+import { round } from "../utilities/utilities";
+
+/**
+ * @typedef {Object} discomfortIndexReturnType
```
szaldivar, line 4: Types usually start with capital letter, so `DiscomfortIndexReturnType`

src/models/discomfort_index.js
```diff
@@ -0,0 +1,98 @@
...
+ * @property {number} di – Discomfort Index(DI)
+ * @property {string} discomfort_condition Classification of the thermal comfort conditions according to the discomfort index
+ */
+
+/**
+ * Calculates the Discomfort Index (DI). The index is essentially an effective temperature based on air temperature and humidity.
+ * The discomfort index is usuallly divided in 6 dicomfort categories and it only applies to warm environments.
+ * @see {@link discomfort_index_array} for a version that supports arrays
+ *
```
szaldivar, line 13: mark as `@public` so it gets added to docs with the config I am making

src/models/discomfort_index.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  return {
+    di: round(di, 1),
+    discomfort_condition: condition,
+  };
+}
+
+/**
+ * @typedef {Object} discomfortIndexArrayReturnType
```
szaldivar, line 48: same comment as before, `Discom...` instead of `discom...`

src/models/discomfort_index.js
```diff
@@ -0,0 +1,98 @@
...
+ * Calculates the Discomfort Index (DI). The index is essentially an effective temperature based on air temperature and humidity.
+ * The discomfort index is usuallly divided in 6 dicomfort categories and it only applies to warm environments.
+ * @see {@link discomfort_index} for a version that supports scalar arguments
+ *
+ * @param {number[]} tdb - air temperature [C]
+ * @param {number[]} rh - relative humidity [%]
+ *
+ * @returns {discomfortIndexArrayReturnType} object with results of DI
+ */
+export function discomfortIndex_array(tdb, rh) {
```
szaldivar, line 63: name is not consistent with the previous `discomfort_index` function

src/models/discomfort_index.js
```diff
@@ -0,0 +1,98 @@
...
+  tdb = Array.isArray(tdb) ? tdb : [tdb];
```
szaldivar, line 64: no need for this checks, you expect them to be an array

src/models/discomfort_index.js
```diff
@@ -0,0 +1,98 @@
...
+  rh = Array.isArray(rh) ? rh : [rh];
+
+  const di = tdb.map((temperature, index) => {
+    const relativeHumidity = rh[index];
+    return (
+      temperature - 0.55 * (1 - 0.01 * relativeHumidity) * (temperature - 14.5)
+    );
+  });
+
+  const diCategories = {
```
szaldivar, line 74: same comment about not using objects like this

src/models/discomfort_index.js
```diff
@@ -0,0 +1,98 @@
...
+    21: "No discomfort",
+    24: "Less than 50% feels discomfort",
+    27: "More than 50% feels discomfort",
+    29: "Most of the population feels discomfort",
+    32: "Everyone feels severe stress",
+    99: "State of medical emergency",
+  };
+
```
szaldivar, line 82: can you implement a function/functions that covers the common code in both `_array` and non array version

tests/index.test.js
```diff
@@ -38,6 +38,10 @@ describe("NPM Package", () => {
     expect(jsthermalcomfort.models).toHaveProperty("wbgt");
   });
 
+  it("should have models.discomfort_index", () => {
+    expect(jsthermalcomfort.models).toHaveProperty("discomfort_index");
```
szaldivar, line 42: missing tests for array version

tests/models/discomfort_index.test.js
```diff
@@ -0,0 +1,86 @@
...
+        discomfort_condition: [
+          "Most of the population feels discomfort",
+          "State of medical emergency",
+          "No discomfort",
+        ],
+      },
+    },
+  ])("returns", ({ tdb, rh, expected }) => {
+    const result = discomfortIndex_array(tdb, rh);
+    expect(result).toStrictEqual(expected, 1);
```
szaldivar, line 84: can you use `deep_close_to_object` or something like that from `test_utilities.js`

szaldivar (changes requested):

src/models/discomfort_index.js
```diff
@@ -0,0 +1,91 @@
+import { round } from "../utilities/utilities";
+
+/**
+ * @typedef {Object} DiscomfortIndexReturnType
+ * @property {number} di – Discomfort Index(DI)
+ * @property {string} discomfort_condition Classification of the thermal comfort conditions according to the discomfort index
```
szaldivar, line 6: do you know if you can specify this return type in-place in the `@returns` tag? it would be nice so it shows nicer in the documentation but I am not sure if JSDoc supports this

szaldivar: what is the answer to this?

src/models/discomfort_index.js
```diff
@@ -0,0 +1,90 @@
...
+ * @param {number[]} tdb - air temperature [C]
+ * @param {number[]} rh - relative humidity [%]
+ *
+ * @returns {DiscomfortIndexArrayReturnType} object with results of DI
+ */
+export function discomfort_index_array(tdb, rh) {
+  const di = tdb.map((temperature, index) => {
+    const relativeHumidity = rh[index];
+    return (
+      temperature - 0.55 * (1 - 0.01 * relativeHumidity) * (temperature - 14.5)
```
szaldivar, line 55: can you abstract this calculation into a function, I see it is repeated in the non-array version as well.
The reason for this is that in the future if someone wants to change the calculation, they might change it only in one place leading to inconsistencies in the non-array vs the array version.
if you put it in a function and someone later changes it then it will take effect for both versions

szaldivar (changes requested):

src/models/discomfort_index.js
YuchenMu: I am not sure if I understand it properly. The update I did is to add explanations for the ReturnType. Can you give me an example of this?

szaldivar: I was just wondering if there is a way to do it inline, so instead of using a type for `@returns` maybe it can be done in line:
```
@returns {object} (?? something here?)
```
so we don't have to include the type in the docs but I don't know if it is possible

src/models/discomfort_index.js
```diff
@@ -0,0 +1,98 @@
...
+ */
+export function discomfort_index_array(tdb, rh) {
+  const di = tdb.map((temperature, index) =>
+    calculate_di(temperature, rh[index]),
+  );
+
+  const discomfortCondition = di.map((value) => check_categories(value));
+
+  return {
+    di: di.map((value) => Math.round(value * 10) / 10),
```
szaldivar, line 59: use round from utilities, like you did above

src/models/discomfort_index.js
YuchenMu: Sorry, I have little experience with that. I'm just trying to learn from you and follow all your guidance :(

src/models/discomfort_index.js
YuchenMu: done!

src/models/discomfort_index.js
YuchenMu: done!

szaldivar (approved):

PR #25, opened 2023-09-06
implement athb & athb_array, test and index included

Lu-Whale:
function athb & athb_array in athb.js

### DIFF
--- a/src/models/athb.js
+++ b/src/models/athb.js
@@ -1,0 +1,256 @@
+/**
+ * Return the PMV value calculated with the Adaptive Thermal Heat Balance
+ * Framework [27]_. The adaptive thermal heat balance (ATHB) framework
+ * introduced a method to account for the three adaptive principals, namely
+ * physiological, behavioral, and psychological adaptation, individually
+ * within existing heat balance models. The objective is a predictive model of
+ * thermal sensation applicable during the design stage or in international
+ * standards without knowing characteristics of future occupants.
+ *
+ * This is a version that supports scalar arguments.
+ * @see {@link athb_array} for a version that supports arrays.
+ *
+ * @param { number } tdb - dry bulb air temperature, in [°C]
+ * @param { number } tr - mean radiant temperature, in [°C]
+ * @param { number } vr - relative air speed, in [m/s]
+ *
+ * Note: vr is the relative air speed caused by body movement and not the air
+ * speed measured by the air speed sensor. The relative air speed is the sum of the
+ * average air speed measured by the sensor plus the activity-generated air speed
+ * (Vag). Where Vag is the activity-generated air speed caused by motion of
+ * individual body parts. vr can be calculated using the function
+ * jsthermalcomfort.utilities.v_relative.
+ *
+ * @param { number } rh - relative humidity, [%]
+ * @param { number } met - metabolic rate, [met]
+ * @param { number } t_running_mean - running mean temperature, in [°C]
+ *
+ * The running mean temperature can be calculated using the function
+ * jsthermalcomfort.utilities.running_mean_outdoor_temperature.
+ *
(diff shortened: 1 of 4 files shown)


src/models/athb.js
```diff
@@ -0,0 +1,256 @@
+/**
+ * Return the PMV value calculated with the Adaptive Thermal Heat Balance
+ * Framework [27]_. The adaptive thermal heat balance (ATHB) framework
+ * introduced a method to account for the three adaptive principals, namely
+ * physiological, behavioral, and psychological adaptation, individually
+ * within existing heat balance models. The objective is a predictive model of
+ * thermal sensation applicable during the design stage or in international
+ * standards without knowing characteristics of future occupants.
+ *
```
szaldivar, line 9: add `@public` so it gets included in the documentatation

src/models/athb.js
```diff
@@ -0,0 +1,256 @@
...
+ * // Import the athb function from the jsthermalcomfort module
+ * import { athb } from "../../src/models/athb.js";
+ *
+ * // Calculate PMV with sample data when it is scalar arguments.
+ * console.log(athb({ tdb: 25, tr: 25, vr: 0.1, rh: 50, met: 1.1, t_running_mean: 20 }));
+ * // Output: 0.2
+ *
+ */
+
```
szaldivar, line 44: remove empty line

src/models/athb.js
```diff
@@ -0,0 +1,256 @@
...
+ * @returns { number } athb_pmv - Predicted Mean Vote calculated with the Adaptive Thermal Heat Balance framework
+ *
+ * @example
+ * // JavaScript code example
+ *
+ * // Import the athb function from the jsthermalcomfort module
+ * import { athb } from "../../src/models/athb.js";
+ *
+ * // Calculate PMV with sample data when it is scalar arguments.
+ * console.log(athb({ tdb: 25, tr: 25, vr: 0.1, rh: 50, met: 1.1, t_running_mean: 20 }));
```
szaldivar, line 40: this example is wrong, in the function you take params but here you are passing an object

src/models/athb.js
```diff
@@ -0,0 +1,256 @@
...
+ *
+ * // Import the athb_array function from the jsthermalcomfort module
+ * import { athb_array } from "../../src/models/athb.js";
+ *
+ * // Calculate PMV with sample data when it is arrays.
+ * console.log(athb({ tdb: [25, 27], tr: [25, 25], vr: [0.1, 0.1], rh: [50, 50], met: [1.1, 1.1], t_running_mean: [20, 20] }));
+ * // Output: [0.2, 0.209]
+ */
+export function athb_array(tdb, tr, vr, rh, met, t_running_mean) {
+  tdb = Array.isArray(tdb) ? tdb : [tdb];
```
szaldivar, line 124: no point in doing these checks, your params are arrays

src/models/athb.js
```diff
@@ -0,0 +1,256 @@
...
+        0.0342 * t_running_mean[index] +
+        0.0002264 * lValue * t_running_mean[index] +
+        0.018696 * met_adapted[index] * t_running_mean[index] -
+        0.0002909 * lValue * met_adapted[index] * t_running_mean[index]
+      ).toFixed(3),
+    ),
+  );
+}
+
+function pmv_ppd_optimized(tdb, tr, vr, rh, met, clo, wme) {
```
szaldivar, line 181: why is this function here? (should not include double implementations in multiple PR)
also why is it called optimized?
what is optimized about it? (aka this does not make sense in the javascript version) (the python version was using a vectorized technique, also using JIT and caching) (here we are not doing any of that so no point in calling it optimized)

szaldivar (changes requested):

src/models/athb.js
Lu-Whale: so here, should I use the function pmv instead?

PR #26, opened 2023-09-07
Better docs

szaldivar:
* Docs with only public functions
* Added handling of custom `@docname` for display names in documentation
* Better table of contents showing all functions.
* Added grouping per functions (models, psychrometrics, utilities)
* Better display of titles and subtitles
* Don't collapse documentation on the content page


szaldivar:
You can ignore most of `docs_theme`, it is the default template for documentation js but with quite a few modifications to improve how we manage and how the docs will look.

src/models/heat_index.js
```diff
@@ -3,6 +3,10 @@ import { round } from "../utilities/utilities";
 /**
  * Calculates the Heat Index (HI). It combines air temperature and relative humidity to determine an apparent temperature.
  *
+ * @public
+ * @memberof models
+ * @docname Heat Index
```
szaldivar, line 8: I added functionality in our theme to support this custom tag and the text after it will be used for the display name of the function (the title in the docs, we can still see the real function name in the docs)

src/psychrometrics/p_sat.js
```diff
@@ -3,6 +3,9 @@ import { round } from "../utilities/utilities.js";
 /**
  * Calculates vapour pressure of water at different temperatures
  *
+ * @public
+ * @memberof psychrometrics
+ *
```
szaldivar, line 8: psychrometrics don't have a title in the Python version, so if we omit `@docname` I coded it so it will default to the function signature

JordanOcokoljic (approved):
Love this - approved!

src/psychrometrics/p_sat.js
JordanOcokoljic: Works for me!

src/models/heat_index.js
JordanOcokoljic: Sick, sounds good

PR #28, opened 2023-09-08
add ref value

Ilite-Kiriya:


src/utilities/utilities.js
```diff
@@ -568,3 +568,38 @@ export function valid_range(range, [min, max]) {
   if (range === undefined) return [];
   return range.map((n) => (n >= min && n <= max ? n : NaN));
 }
+
+// Met values of typical tasks.
```
szaldivar, line 572: Can you include JSDoc comments, please? [see here](https://pythermalcomfort.readthedocs.io/en/latest/reference/pythermalcomfort.html#reference-values-clo-and-met)
Remember to add `@public` `@docname` and `@memberof` so it shows correctly in the docs

szaldivar (changes requested):

src/utilities/utilities.js
```diff
@@ -568,3 +568,79 @@ export function valid_range(range, [min, max]) {
   if (range === undefined) return [];
   return range.map((n) => (n >= min && n <= max ? n : NaN));
 }
+
+/**
+ * Met values of typical tasks.
+ * 
+ * @public
+ * @memberof Reference_values_clo_and_met
```
szaldivar, line 576: You will need to create than name, also better if it is called `reference_values`
You can create the name in `utilities/index.js` above the comment that has `@name utilities` (put it in a separated jsdoc comment)

szaldivar (changes requested):

src/utilities/utilities.js
```diff
@@ -568,3 +568,79 @@ export function valid_range(range, [min, max]) {
...
+ * @docname Met typical tasks, [met]
+ * 
+ * @constant met_typical_tasks - {
```
szaldivar, line 579: What happens if you just put,`@constant` without all that text, hopefully it gets rendered from the value actually declared in the export rather than this comment

szaldivar (changes requested):

src/utilities/index.js
```diff
@@ -7,14 +7,15 @@ import {
...
   f_svv,
+  met_typical_tasks,
 } from "./utilities";
 
 /**
  * @public
  * @name utilities
  * @docname Utilities
  */
-export default {
```
szaldivar, line 17: Don't remove this

src/utilities/index.js
```diff
@@ -24,3 +25,13 @@ export default {
...
 };
+
+/**
+ * @public
+ * @name reference_values
+ * @docname Reference values clo and met
+ */
+
+export {
+  met_typical_tasks
```
szaldivar, line 36: Export it in the default export

src/utilities/index.js
```diff
@@ -24,3 +25,13 @@ export default {
   running_mean_outdoor_temperature,
   f_svv,
 };
+
+/**
+ * @public
+ * @name reference_values
```
szaldivar, line 31: This jsdoc comment can live on its own, don't need to separate the exports

szaldivar (changes requested):

Ilite-Kiriya:
Don't review it. I am still working on it. : ) Just push to remote repo for a second copy

szaldivar:
No worries, I think you can mark this as draft if that is the case

PR #29, opened 2023-09-08
Custom types documentation

szaldivar:
- Custom types used in functions are now injected into the function block so they get rendered together.
- Only need to add  a`@public` tag to the type definition and the custom documentation generator will handle the rest with this PR.
- ~~Added support for properties with spaces.~~ (removed this since it does not work too well with IDEs, we will replace spaces with underscores)
- Added support for references. Only need to type `{@link /#ref_<num>|[<num>]}` where you want a reference and add the reference to `docs_theme/references.md` on a new line and ordered (look at the format of other references in there, it is important for the support I added)


JordanOcokoljic (approved):
I'm all good with this!

PR #31, opened 2023-09-09
Test that docs can build on PRs

szaldivar:
- This will make sure that we do not forget to include the public types used by a function and that we can handle the syntax in references.md


JordanOcokoljic (approved):
Excellent!

PR #34, opened 2023-09-09
pmv_ppd and pmv_ppd_array

Lu-Whale:
pmv_ppd implemented


szaldivar:
SET tmp is now in main, can you resolve the conflicts and make sure that the tests pass. I will review it after that

szaldivar:
Raise the PR for cooling effect since now set tmp is in main

Lu-Whale:
> Raise the PR for cooling effect since now set tmp is in main

No problem, I will do that this weekend

szaldivar:
now that colling effect is in you can probably finish this one

src/models/pmv_ppd.js
```diff
@@ -0,0 +1,486 @@
...
+ * this is then subtracted from both the air and mean radiant temperatures, and the
+ * differences are used as input to the PMV model, while the airspeed is set to 0.1m/s.
+ * Please read more in the Note below.
+ *
+ * Notes:
+ * You can use this function to calculate the `PMV` and `PPD` in accordance with
+ * either the ASHRAE 55 2020 Standard {@link #ref_1|[1]} or the ISO 7730 Standard {@link #ref_2|[2]}.
+ *
+ * _PMV:
+ * {@link https://en.wikipedia.org/wiki/Thermal_comfort#PMV/PPD_method|PMV}
```
szaldivar, line 48: Maybe we can put these links in the paragraph? so instead of `PMV` you could do what you are doing here

src/models/pmv_ppd.js
```diff
@@ -0,0 +1,486 @@
...
+ * @param { "ISO"|"ASHRAE" } [standard="ISO"] - comfort standard used for calculation
+ * - If "ISO", then the ISO Equation is used
+ * - If "ASHRAE", then the ASHRAE Equation is used
+ * Note: While the PMV equation is the same for both the ISO and ASHRAE standards,
+ * the ASHRAE Standard Use of the PMV model is limited to air speeds below 0.10m/s (20 fpm).
+ * When air speeds exceed 0.10 m/s (20 fpm), the comfort zone boundaries are adjusted based on the SET model.
+ * This change was introduced by the `Addendum C to Standard 55-2020`_
+ * @param { Pmv_ppdKwargs }kwargs - additional arguments
+ *
+ * @returns { Object } - Result of pmv and ppd
```
szaldivar, line 88: I don't think this works, you have to define a `@typedef` type and return that with the two properties below

src/models/pmv_ppd.js
```diff
@@ -0,0 +1,486 @@
...
+    airspeed_control: true,
+  };
+  kwargs = Object.assign(default_kwargs, kwargs);
+
+  if (kwargs.units && kwargs.units === "IP") {
+    // Conversion from IP to SI units
+    const result = units_converter({ tdb, tr, vr }, "IP");
+    tdb = result.tdb;
+    tr = result.tr;
+    vr = result.vr;
```
szaldivar, line 130: this is fine, but just so you know you can use the following syntax to do the same:
```javascript
({tdb, tr, vr} = units_converter({ tdb, tr, vr }, "IP"));
```

src/models/pmv_ppd.js
```diff
@@ -0,0 +1,486 @@
...
+  });
+
+  // Checks that inputs are within the bounds accepted by the model if not return NaN
+  if (kwargs.limit_inputs) {
+    const pmv_valid =
+      standard === "ASHRAE"
+        ? valid_range(pmv_array, [-100, 100])
+        : valid_range(pmv_array, [-2, 2]); // this is the ISO limit
+
+    const all_valid = pmv_array.map((valid, i) => {
```
szaldivar, line 377: you don't need this right? just check them in the for loop below so you only walk the arrays once rather than twice

src/models/pmv_ppd.js
```diff
@@ -0,0 +1,486 @@
...
+        isNaN(tr_valid[i]) ||
+        isNaN(v_valid[i]) ||
+        isNaN(met_valid[i]) ||
+        isNaN(clo_valid[i]) ||
+        isNaN(pmv_valid[i])
+      );
+    });
+
+    for (let i = 0; i < pmv_array.length; i++) {
+      if (!all_valid[i]) {
```
szaldivar, line 390: do the `isNaN(tdb_valid[i])` for all the arrays here and remove the `all_valid` computation so you only do this work once

src/models/pmv_ppd.js
```diff
@@ -0,0 +1,486 @@
...
+        pmv_array[i] = NaN;
+        ppd_array[i] = NaN;
+      }
+    }
+  }
+
+  return {
+    pmv: pmv_array.map((value) => round(value, 2)),
```
szaldivar, line 398: can you do this in place rather than creating a new array? (so we avoid allocating more memory)
something like:
```javascript
for (let i = 0; i < pmv_array.length; ++i) {
  pmv_array[i] = round(pmv_array[i], 2)
}
```

szaldivar (changes requested):

src/models/pmv_ppd.js
Lu-Whale: done

src/models/pmv_ppd.js
Lu-Whale: done

src/models/pmv_ppd.js
Lu-Whale: done

src/models/pmv_ppd.js
Lu-Whale: done

src/models/pmv_ppd.js
Lu-Whale: done

src/models/pmv_ppd.js
Lu-Whale: done

szaldivar (approved):

PR #35, opened 2023-09-10
Add `clo_tout` and `clo_tout_array`

JordanOcokoljic:


docs/index.html
szaldivar, line 1: might want to merge with main, commit, then regen and commit again (so github links work correctly)

src/models/index.js
```diff
@@ -16,7 +17,11 @@ export default {
   phs,
   humidex,
   net,
+  clo_tout,
+  clo_tout_array,
   wbgt,
+  clo_tout,
+  clo_tout_array,
```
szaldivar, line 24: adding these functions twice?

docs/index.html
JordanOcokoljic: Will do

src/models/index.js
JordanOcokoljic: Ah the merge was _very_ ugly, I'll fix that

szaldivar (approved):

PR #39, opened 2023-09-10
Yuchen/two nodes new api

YuchenMu:
two_nodes float and array version new API


src/models/index.js
```diff
@@ -4,6 +4,8 @@ import { humidex } from "./humidex";
 import { net } from "./net";
 import { wbgt } from "./wbgt";
 import { discomfort_index, discomfort_index_array } from "./discomfort_index";
+
```
szaldivar, line 7: can you remove the empty line please?

src/models/two_nodes.js
```diff
@@ -0,0 +1,886 @@
+import { round } from "../utilities/utilities.js";
+import { p_sat_torr } from "../psychrometrics/p_sat_torr.js";
+
+/**
+ * @typedef {Object} TwoNodesReturnType
```
szaldivar, line 5: for the props in this type please follow the same naming conventions as the python library:
so `eSkin` -> `e_skin` etc

src/models/two_nodes.js
```diff
@@ -0,0 +1,886 @@
...
+ * negative category descriptions used for Fanger’s PMV apply.
+ * Heat gains and losses via convection, radiation and conduction.
+ * The Standard Effective Temperature (SET)
+ * The New Effective Temperature (ET)
+ * The Predicted Thermal Sensation (TSENS)
+ * The Predicted Percent Dissatisfied Due to Draft (PD)
+ * Predicted Percent Satisfied With the Level of Air Movement” (PS)
+ * 
+ * @public
+ * @memberof models
```
szaldivar, line 70: add `@docname` and the public name of the function for the docs (Gagge et al. two-node model)

src/models/two_nodes.js
```diff
@@ -0,0 +1,886 @@
...
+
+/**
+ * Round all elments in a given array to the given precision.
+ *
+ * @param {number[]} array - the array to round
+ * @param {number} precision - the number of decimal places to round to
+ * @returns the rounded result
+ */
+function roundArray(array, precision) {
+  return array.map((value) => round(value, precision));
```
szaldivar, line 412: can you do this in place rather than creating a new array?

src/models/two_nodes.js
```diff
@@ -0,0 +1,886 @@
...
+ *
+ */
+export function two_nodes(
+  tdb,
+  tr,
+  v,
+  rh,
+  met,
+  clo,
+  wme = 0,
```
szaldivar, line 120: if these parameters have default values mark them as such in the JSDoc

src/models/two_nodes.js
```diff
@@ -0,0 +1,886 @@
...
+    w_max: false,
+  };
+
+  let joint_kwargs = Object.assign(defaults_kwargs, kwargs);
+
+  const vaporPressureArray = rhArray.map((_rh, index) =>
+    cal_vapor_pressure(tdbArray[index], _rh),
+  );
+  const wmeArray = fill_array(wme, tdbArray);
+  const bodySurfaceArray = fill_array(body_surface_area, tdbArray);
```
szaldivar, line 301: I rather you not do this. This would hide erroneous input from the user.
Better to don't give them a default value in the signature and check if they are undefined, if they are then you create the arrays with default values

src/models/two_nodes.js
```diff
@@ -0,0 +1,886 @@
...
+  const pAtmArray = fill_array(p_atmospheric, tdbArray);
+  const bodyPositionArray = fill_array(body_position, tdbArray);
+  const maxSkinBloodFlowArray = fill_array(max_skin_blood_flow, tdbArray);
+
+  const setArray = [];
```
szaldivar, line 306: build the arrays to the correct size here so you avoid having to copy elements when you container needs to resize in the loop

src/models/two_nodes.js
```diff
@@ -0,0 +1,886 @@
...
+      metArray[index],
+      cloArray[index],
+      vaporPressureArray[index],
+      wmeArray[index],
+      bodySurfaceArray[index],
+      pAtmArray[index],
+      bodyPositionArray[index],
+      joint_kwargs,
+    );
+    setArray.push(result.set);
```
szaldivar, line 339: see the previous comment, if you allocate the arrays before hand you can just set the values to the index here, avoiding pushing and causing the array to reallocate and copy values over

szaldivar (changes requested):

src/models/two_nodes.js
```diff
@@ -186,32 +187,32 @@ const skinBloodFlowNeutral = 6.3;
...
+ * @property {number[]} t_sens  – Array of Predicted Thermal Sensation
  * @public
  */
 
 /**
  * @typedef {Object} TwoNodesArrayKwargs
  * @property {boolean} [round=true]
  * @property {boolean} [calculate_ce=false]
- * @property {number[]} [max_sweating]
+ * @property {number[]} [max_sweating=500]
```
szaldivar, line 215:
```suggestion
 * @property {number[]} [max_sweating]
```

szaldivar (approved):

PR #45, opened 2023-09-11
finish transpose sharp altitude

Ilite-Kiriya:


src/utilities/utilities.js
```diff
@@ -40,6 +40,33 @@ export function round(number, precision) {
  * @typedef {"ankle_draft" | "ashrae" | "iso" | "ISO7933"} Standard
  */
 
+function degrees_to_radians(degrees) {
+  var pi = Math.PI;
+  return degrees * (pi / 180);
```
szaldivar, line 45: just use `Math.PI` in the calculation, no need for a variable. Same in the other function

src/utilities/utilities.js
```diff
@@ -40,6 +40,33 @@ export function round(number, precision) {
...
+}
+
+function radians_to_degrees(radians) {
+  var pi = Math.PI;
+  return radians * (180 / pi);
+}
+//There is no documentation of this function
```
szaldivar, line 52: document the types at least

src/utilities/utilities.js
```diff
@@ -40,6 +40,33 @@ export function round(number, precision) {
  * @typedef {"ankle_draft" | "ashrae" | "iso" | "ISO7933"} Standard
  */
 
+function degrees_to_radians(degrees) {
```
szaldivar, line 43: add types to functions

szaldivar (changes requested):

src/utilities/utilities.js
```diff
@@ -40,6 +40,50 @@ export function round(number, precision) {
...
+ */
+function radians_to_degrees(radians) {
+  return radians * (180 / Math.PI);
+}
+
+/**
+ * Converts sharp and altittude from radians to degree unit
+ * @param {number} sharp
+ * @param {number} altitude
+ * @returns {number}
```
szaldivar, line 69: this function returns a tuple right? 
```suggestion
 * @returns {[number, number]}
```

szaldivar (changes requested):

PR #53, opened 2023-09-16
Yuchen/doc fix

YuchenMu:
Update some docs in two_nodes and set_tmp according to Omar's request


src/models/two_nodes.js
```diff
@@ -256,7 +257,29 @@ const skinBloodFlowNeutral = 6.3;
 
 /**
  * 
- * Compute two nodes model when the input parameters are arrays. 
+ * Two nodes model of human temperature regulation Gagge et al. when the input parameters are arrays. 
+ * 
+ * {@link #ref_25|[25]} This model can be used to calculate a variety of indices, including:
+ * 
+ *  - Gagge’s version of Fanger’s Predicted Mean Vote (PMV). This function uses
```
szaldivar, line 264: why is the format here and the non-array version different? I think I prefer this format

szaldivar (changes requested):

szaldivar (approved):

PR #72, opened 2023-09-21
basic setup for releasing

szaldivar:
- Adding back the tsconfigs to generate ESM and CJS ways to export with their types from JSDocs
- Changing imports to use `.js` so it works on the web
- Adding a GitHub action to check that we can build the project


Lu-Whale (approved):
AWESOME

PR #78, opened 2023-09-25
Use fans heatwave finished

YuchenMu:


src/models/use_fans_heatwave.js
```diff
@@ -0,0 +1,253 @@
...
+ * @param {number} max_skin_blood_flow - maximum blood flow from the core to the skin, [kg/h/m2] default 80
+ * @returns {object} heat strain conditions
+ */
+function cal_heatwave(two_nodes_result, joint_kwargs, max_skin_blood_flow) {
+  let heat_strain_blood_flow,
+    heat_strain_w,
+    heat_strain_sweating,
+    heat_strain = 0;
+
+  heat_strain_blood_flow = two_nodes_result.mBl === max_skin_blood_flow ? 1 : 0;
```
szaldivar, line 236: why are these variables numbers instead of booleans?
also, these comparisons are float comparisons, right? if that is the case you need to do close to rather than a comparison

szaldivar (changes requested):

src/models/use_fans_heatwave.js
```diff
@@ -231,19 +240,18 @@ function cal_heatwave(two_nodes_result, joint_kwargs, max_skin_blood_flow) {
   let heat_strain_blood_flow,
     heat_strain_w,
     heat_strain_sweating,
-    heat_strain = 0;
+    heat_strain = false;
 
-  heat_strain_blood_flow = two_nodes_result.mBl === max_skin_blood_flow ? 1 : 0;
-  heat_strain_w = two_nodes_result.w === two_nodes_result.wMax ? 1 : 0;
+  heat_strain_blood_flow =
+    two_nodes_result.mBl === max_skin_blood_flow ? true : false;
```
szaldivar, line 246: you can remove the `? true : false;` part for all of these, it is redundant

szaldivar: I have the feeling that this should be a `>=` since it is a max

szaldivar: can you confirm in the python code?

szaldivar (changes requested):

szaldivar (approved):

PR #80, opened 2023-10-01
utci finished

Ilite-Kiriya:


src/models/utci.js
```diff
@@ -0,0 +1,619 @@
+import { units_converter, valid_range, round } from "../utilities/utilities.js";
+
+const stress_categories = {
```
szaldivar, line 3: Better to use an array and do a binary search over it. @YuchenMu had a similar implementation for this. you can talk with him in how to do this in a better way

src/models/utci.js
```diff
@@ -0,0 +1,619 @@
...
+
+  const eh_pa = exponential(tdb) * (rh / 100.0);
+  const delta_t_tr = tr - tdb;
+  const pa = eh_pa / 10.0; // convert vapour pressure to kPa
+
+  let utci_approx = utci_optimized(tdb, v, delta_t_tr, pa);
+
+  // Checks that inputs are within the bounds accepted by the model if not return nan
+  if (limit_inputs) {
+    const tdb_valid = valid_range(tdb, [-50.0, 50.0]);
```
szaldivar, line 78: valid_range is for arrays, just do a normal `tdb >= -50.0 && tdb <=50.0` here

src/models/utci.js
```diff
@@ -0,0 +1,619 @@
...
+        isNaN(v_valid[i])
+      );
+      if (!all_valid) {
+        utci_approx[i] = NaN;
+      }
+    }
+  }
+
+  if (units.toLowerCase() == "ip") {
+    for (let i = 0; i < utci_approx.length; i++) {
```
szaldivar, line 175: I think we have a `units_converter` for arrays

src/models/utci.js
```diff
@@ -0,0 +1,619 @@
...
+  return ret;
+}
+
+/**
+ *
+ * @param {number} t_db
+ * @returns {number}
+ */
+function exponential(t_db) {
+  const g = [
```
szaldivar, line 227: can you declare this outside the function since it is a constant so we don't allocate this memory every time

src/models/utci.js
```diff
@@ -0,0 +1,619 @@
...
+    -6028.076559,
+    19.54263612,
+    -0.02737830188,
+    0.000016261698,
+    7.0229056 * Math.pow(10.0, -10),
+    -1.8680009 * Math.pow(10.0, -13),
+  ];
+  const tk = t_db + 273.15; // air temp in K
+  let es = 2.7150305 * Math.log1p(tk);
+  g.map((_g, i) => {
```
szaldivar, line 238: do a for loop. map creates an array so if you don't need to create an array don't use map

src/models/utci.js
```diff
@@ -0,0 +1,619 @@
...
+}
+
+// tdb,
+//   tr,
+//   v,
+//   rh,
+//   units = "SI",
+//   return_stress_category = false,
+//   limit_inputs = true,
+// console.log(utci_array([25, 25], [27, 25], [1, 1], [50, 50], "si", true))
```
szaldivar, line 619: ?

src/utilities/utilities.js
```diff
@@ -713,6 +713,9 @@ export function f_svv(w, h, d) {
  * @returns {number[]} the constrained range with NaNs for values that are outside the min, max range
  */
 export function valid_range(range, [min, max]) {
+  if (!Array.isArray(range)) {
```
szaldivar, line 716: don't do this, if you need a numeric comparison just do the numeric comparison

szaldivar (changes requested):

src/models/utci.js
Ilite-Kiriya: will delete it later

szaldivar (approved):

PR #97, opened 2023-10-10
add vasomotricity and sweat_rate to pet_steady.js

Ilite-Kiriya:


src/models/pet_steady.js
```diff
@@ -0,0 +1,59 @@
...
+
+// skin and core temperatures set values
+const tc_set = 36.6;
+const tsk_set = 34;
+
+/**
+ * Defines the vasomotricity (blood flow) in function of the core and skin temperatures.
+ * @param {number} t_cr - The body core temperature, [°C]
+ * @param {number} t_sk - The body skin temperature, [°C]
+ * @returns {Object} m_blood - Blood flow rate, [kg/m2/h]
```
szaldivar, line 12: create a custom type for this with the properties so we get code completion

szaldivar (changes requested):

src/models/pet_steady.js
```diff
@@ -0,0 +1,65 @@
+import { body_surface_area } from "../utilities/utilities";
+import { p_sat } from "../psychrometrics/p_sat";
+
+/**
+ * @public
```
szaldivar, line 5:
```suggestion
```
does not need to be public since `vasomotricity` is not public

szaldivar (changes requested):

szaldivar (approved):

PR #109, opened 2023-10-17
Add thermoregulation functions

JordanOcokoljic:


szaldivar:
I think the only need is to change to imports from "something" to "something.js" so they work in the browser

JordanOcokoljic:
@szaldivar fixed!

PR #110, opened 2023-10-22
NPM publish on GitHub releases

szaldivar:
- @FedericoTartarini This requires the repository to have a secret called `NPM_TOKEN` with the NPM token.
- This PR will make it so that when we create a release using the GitHub release system, it will also trigger a publish to NPM. (I like this pattern since it means that we will see all the releases in the repository for quick access, and they will be synced with NPM releases.)
- Bumping the version to `0.1.0`

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,18 @@
+name: Publish Package to NPM
+on:
+  release:
+    types: [published]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


Lu-Whale (approved):
LGTM

FedericoTartarini:
Sure this is a great idea. I have already added the NPM_TOEKEN environmental variable to the project. Is there anything else I need to do?

szaldivar:
Great! I don't think there is anything else. I will test it out and let you know

PR #120, opened 2023-11-04
Use .js for all imports

szaldivar:


JordanOcokoljic (approved):
Good catch, approved

YuchenMu (approved):
Thank you for your work!

PR #125, opened 2024-02-02
Fix issues with JOS3, migrate most functions to use `math.Matrix` instead of plain arrays

JordanOcokoljic:
This PR fixes a standing issue in JOS3 that would prevent it from running more than one simulation run.

As a consequence, many of the functions that JOS3 relies on have been updated to accept and return `mathjs` `Matrix` instances, as they provide much better flexibility for common mathematical operations.

Therefore, the next publication of the library should be to a 2.x version.


JordanOcokoljic:
On checking, the library is currently published under `v0.x` which means breaking changes are _tolerable_ - so no need to worry with bump candidacy. To indicate the changes are significant I'm proposing that we'll move to `v0.2.0` from `v0.1.5`

FedericoTartarini (approved):
Fantastic work, thank you so much for fixing the errors and updating the code

PR #126, opened 2024-04-22
Run build in prep for NPM release

JordanOcokoljic:
This should have been done with the last commit, but it slipped my mind when pushing it. 

There are no changes to actual code, just the built `/lib` directory that provides the types to projects that use this - the review here is a formality more than anything else. 

Once this is in, we can trigger a new release via NPM so that our users get all the latest and greatest from this project.


FedericoTartarini (approved):

FedericoTartarini:
Thank you so much for working on this. Please trigger a new release of the NPM package whenever you have time. Where you able to fix all the issues that you mentioned with the JOS3 model?

JordanOcokoljic:
Yes all of those issues got fixed in the last PR I submitted - this one is just because I realised that those changes hadn't managed to work their way through into the package published on NPM.

PR #129, opened 2025-03-17
Enhance test structure and result validation

rcqz:
Refactor the test suite to correctly fetch tolerance and expected output, separate test cases, and introduce a utility for validating results, using `a_pmv_test.js` as an example. 

The test for array input does not pass at the moment because the array case is handled in a separate function `a_pmv_array()` and we are only calling `a_pmv()` in the test cases.


FedericoTartarini (commented):
Great work, I left a few comments, let's discuss them on Friday

rcqz:
I've made some changes to `a_pmv.test.js` and `adaptive_ashrae.test.js`, as well as the functions in `testUtils.js`. 

One side note - in JavaScript, to specify parameter names when calling a function, there's only a workaround by passing in an object. So for example, instead of defining 
`export async function loadTestData(url, returnArray = false)`, 
we would have to do something like 
`export async function loadTestData({ url, returnArray = false })`.

Would you prefer defining everything this way? If so I can reflect this change in my next update

tests/models/a_pmv.test.js
```diff
@@ -1,32 +1,31 @@
...
 
-beforeAll(async () => {
-  const result = await loadTestData(testDataUrls.aPmv, 'pmv');
-  testData = result.testData;
-  tolerance = result.tolerance;
-});
+// use top-level await to load test data before tests are defined.
+let { testData, tolerances } = await loadTestData(
+  testDataUrls.aPmv,
+  returnArray,
```
FedericoTartarini, line 11: I am a bit confused, here you are not returning the array data from the validation table, however, in line 21 you check if the inputs have arrays, this is not consistent. Am I wrong?

FedericoTartarini (commented):
Great work. I left one comment in the test a_pmv function. Please address that then I will accept the Pull request.

tests/models/a_pmv.test.js
rcqz: At the moment we are ignoring all test cases with array inputs, but I wrote it this way in case we want to test the cases with arrays in the futur, just like how `validateResult` handles arrays too. 

Woudl you prefer if we completely ignore arrays for now?

rcqz:
I've fixed the `a_pmv` test so that it ignores tests with array inputs now.

Also there's a problem with one of the `adaptive_ashrae` test cases, where it has `"units": "ip"`. I think this is inconsistent across the test files as some are upper and some are lower cases.

FedericoTartarini (approved):

FedericoTartarini:
@rcqz I have accepted the pull request, however, some of the flows and checks above did not pass, could you please look into it and fix them so we can then merge the pull request? I am aware that some of the tests may fail since you changed the function, but at least the formatting check should pass.

rcqz:
I've just run the formatter on all the files and I think because of the files I haven't got to yet. Some tests are still failing now but I'll try to fix them once i get to those files.

PR #132, opened 2025-04-01
Refactor model return values and improve documentation

rcqz:
Enhance model functions to return structured results instead of single number. Update tests to validate these changes and improve documentation for clarity and consistency.


FedericoTartarini (approved):

FedericoTartarini:
Great work, please look into the comments from Copilot, I will then accept the Pull request.

image

rcqz:
I've now fixed the issues with formatting. Also the `if (isNaN(exp)) {` change is intended because otherwise the if block will always be true when actual value is NaN.

FedericoTartarini (approved):

FedericoTartarini:
@rcqz could you please look into git-pre-commit and add prettier there? So formatting issues will be a thing of the past. https://prettier.io/docs/precommit.html

PR #133, opened 2025-04-06
Refactored tests 

rcqz:
This PR contains most of the tests that are passing after refactoring. The following tests are not updated yet because some of them are failing. 

- `pet_steady.test.js` (2 cases)
- `phs.test.js` (all failing)
- `set_tmp.js` (5 cases)
	- some input values are invalid
- `utci.test.js` (2 cases)
	- these test cases have no input parameter `return_stress_category`, but the expected output contains the `stress_category`
- `pmv.test.js` (validation table does not exist from the given url)

### DIFF
--- a/lib/cjs/types/models/cooling_effect.d.ts
+++ b/lib/cjs/types/models/cooling_effect.d.ts
@@ -1,4 +1,9 @@
 /**
+ * @typedef {object} CoolingEffectResult
+ * @property {number} ce - Cooling Effect, default in [°C] in [°F] if `units` = 'IP'
+ * @public
+ */
+/**
  * Returns the value of the Cooling Effect ( {@link https://en.wikipedia.org/wiki/Thermal_comfort#Cooling_Effect|CE} )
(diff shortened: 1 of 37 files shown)


rcqz:
I've tried to fix most of the test cases, but there are still three of them that do not work after trying the feedback I got on Monday:
1. `pet_steady.test.js`: Only the last 2 test cases are failing. I've checked the default posture but I dont think that is the cause. 
2. `phs.test.js`: All test cases are failing. I checked the difference in changing `met` but it does't seem to solve the problem. 
3. `two_nodes.test.js`: 2 of the test cases are failing - I'm not exactly sure why this happens though.

FedericoTartarini (approved):

FedericoTartarini:
I have accepted the pull request.
@rcqz could you please do the following?
1. Bump the version (patch). For example, if the version now is 0.1.0, it should then be 0.1.1. Just update the last value. This process should not be done manually, please read the documentation on how to do it.
1. Deploy a new version of jsthermalcomfort - instructions are in the documentation - if they are not clear plase edit them
2. Deploy the documentation - please check that the documentation is rendering well.

PR #140, opened 2026-03-29
test: add snapshot-based scalar tests for a_pmv, adaptive_ashrae, ada…

JitaoFu:
## What this PR does

- Adds a snapshot-based scalar test framework using `generate_snapshots.py` and `generate_reference_value.py` 
- Adds scalar test cases for `a_pmv`, `adaptive_ashrae`, `adaptive_en`, `ankle_draft`, and `at`
- Adds `tests/snapshots/` directory containing snapshot JSON files with expected outputs generated from pythermalcomfort

## Why this change is needed

- Existing tests only cover validation data from the remote JSON files
- Scalar tests provide additional coverage for edge cases and out-of-range inputs

## How I tested it

- [x] npm test

## Notes for reviewers

- In this PR, "snapshot" refers to a JSON file containing pre-generated expected outputs
from pythermalcomfort (the Python reference implementation). These files are produced
once by running `python scripts/python/generate_snapshots.py` locally, then committed
to the repository.
- When `npm test` runs, each model's test file reads its snapshot JSON and compares the
JS output against the saved expected values. This means CI does not need Python or
pythermalcomfort installed — it only reads the committed JSON files.
- Snapshot JSON files in `tests/snapshots/` must be committed to the repo for CI to pass.
- Tolerances are loaded dynamically from validation data URLs, consistent with existing tests.

See `SNAPSHOT_TESTING.md` for full details on how to add tests for new models.
[SNAPSHOT_TESTING.md](https://github.com/user-attachments/files/26328094/SNAPSHOT_TESTING.md)


FedericoTartarini (commented):
Let's discuss this today but I'm not entirely sure that we should be using this approach. First of all I don't want to be that the JavaScript package is dependent on the Python package. The second thing is that all these snapshots as a JSON file are very easy, difficult to read and manage. I know we are using JSON for the validation table, but that is mostly because we want to share tests across different packages. For the additional tests that we want to include here, we can just write simple tests that we can copy and paste over from Python. I don't think we need to over-engineer this solution.

PR #141, opened 2026-03-29
Remove all *_array function exports

TIAN-TOM:
## What this PR does

- Removes all 18 publicly exported `*_array` functions from `models`, `utilities`, and `psychrometrics`
- Cleans up related imports, JSDoc `@see` references, and test cases across 25 files (−2047 lines)
- Adds migration notes to docs_theme/changelog.md​

## Why this change is needed

- The `*_array` functions were ported from `pythermalcomfort`, where NumPy/Numba makes batch processing efficient. The JS library has no equivalent optimisation, so they provide no performance benefit over a simple `Array.map()` call
- Frontend only uses single-value calculations, so batch APIs are not needed
- Removing them reduces maintenance overhead and simplifies the public API

## How I tested it

- [x] npm test — 67 suites, 1848 tests, 0 failures (baseline was 1880; the 32 removed tests are the deleted `_array` test cases)
- [x] npm run build
- [x] Verified no scalar function behaviour changed by comparing test results before and after

## Notes for reviewers

- `check_standard_compliance_array` is kept — not publicly exported via `index.js`, but used internally by 4 scalar functions (`adaptive_ashrae`, `pmv_ppd`, `set_tmp`, `use_fans_heatwave`)
- `t_o_array` implementation is also kept (not exported) since `check_standard_compliance_array` calls it internally
- `units_converter_array` was fully removed — no implementation or callers remain in `src/`; it was only used by removed `_array` APIs


src/models/adaptive_en.js
```diff
@@ -1,10 +1,5 @@
-import { t_o, t_o_array } from "../psychrometrics/t_o.js";
-import {
-  units_converter,
-  round,
-  units_converter_array,
-  valid_range,
-} from "../utilities/utilities.js";
+import { t_o } from "../psychrometrics/t_o.js";
+import { units_converter, round } from "../utilities/utilities.js";
```
FedericoTartarini: I saw that you removed the function valid_range from the import which makes me a little bit worried because we should be checking that the inputs are within the valid range of applicability of the model. We can either fix it here or we can open a new issue to make sure that the valid range is checked across all the models. Please let me know how you would like to proceed. I think perhaps a new issue is the best approach. So we can check that all the models have this check.

README.md
```diff
@@ -92,3 +92,42 @@ By default tests use the `main` branch. To pin tests to a specific ref (tag/bran
 ```bash
 VALIDATION_DATA_REF=main npm test
 ```
+
```
FedericoTartarini: Please remove this information from here and add it to the changelog file. This should not be in the readme.md

FedericoTartarini (changes requested):
Great work! I left two very minor comments that are going to be very easy for you to implement. One is to remove the instructions to the change log, they should not be in the readme. And the second one is that I would like you to open an issue to make sure that we check the valid range in all the models. and we should add it back to the adaptive model but this could be a separate issue we can first merge this one and then you can work on that separately.

Issue #142, opened 2026-03-30
Add input validation and range checks across all models


TIAN-TOM:
## Problem

Most model functions in jsthermalcomfort currently lack input validation. Passing a string as temperature or a value of 1,000,000 degrees will not raise an error and will return incorrect results. Some functions have partial checks via `check_standard_compliance`, but coverage is incomplete.

## Proposed solution

Following the approach in pythermalcomfort's `limit_inputs` parameter:

- **Type checks:** ensure all numeric parameters are of type `number`
- **Range checks:** return `NaN` when inputs fall outside the documented valid range for each model (e.g. PMV per ASHRAE 55: 10 ≤ tdb < 40, 10 < tr < 40, 0 < vr < 2, 1 < met < 4, 0 < clo < 1.5)

## Scope

All exported model functions in `src/models/`. Can be implemented incrementally — start with functions that already have partial validation and extend to the rest.

## References

- pythermalcomfort `limit_inputs` docs: https://pythermalcomfort.readthedocs.io/en/latest/
- Discussion in team meeting 2026-03-30
- Related: PR #141 review comment by @FedericoTartarini

yzq124537:
I'd like to help with this as well. I can take on the input validation logic for the following models:
- a_pmv
- ankle_draft
- at
- adaptive_en
- adaptive_ashrae

I will follow the same approach using type and range checks as proposed.

yzq124537:
I've put together a validation approach for the models I was assigned — feel free to take a look at my PR (#145) and let me know what you think. If there's anything you'd like to change or discuss, I'm happy to revise it before it gets merged. 

I found some existing range checking functions in the utilities file — check_standard_compliance and check_standard_compliance_array — that might be useful for your work as well.

TIAN-TOM:
Opened PR #146 covering type validation for all 25 exported functions. This is the first half of this issue — range checks with a `limit_inputs` parameter will follow in a separate PR.

JitaoFu:
## Range Check Parity Review
I compared the JS range checks against the Python reference implementation and found a few gaps

#### 1. pmv_ppd
The JS version adds a valid_range(pmv, [-100, 100]) output range check that doesn't exist in the Python implementation. This is extra behaviour not present in the reference — is this intentional?

#### 2. phs (multiple issues)
[...]

PR #145, opened 2026-04-11
feat: implement input parameter type check logic for all models

JitaoFu:
## What this PR does
- Add `validateInputs()` utility function to `utilities.js` supporting type validation for numbers, booleans, enums, and optional fields
- Add input validation logic to all models
- Add corresponding validation tests for all models

## Why this change is needed
- The JS implementation lacked input validation that exists in the Python pythermalcomfort reference implementation. Adding input validation brings the two implementations closer in behaviour, catches invalid inputs early with clear error messages, and reduces the risk of silent failures caused by passing wrong types or invalid values to thermal comfort models.

## How I tested it
- [x] add new tests and run npm test

## Notes for reviewers
None


TIAN-TOM:
Hi @yzq124537, I've just opened PR #146 which also adds input type validation, covering all 25 exported functions. We overlap on `a_pmv`, `adaptive_ashrae`, `adaptive_en`, `ankle_draft`, and `at`. The main difference is that mine returns `NaN` for invalid inputs, while yours throws errors (`TypeError`/`Error`, depending on the field). Might be worth asking Federico which behaviour he wants for the library.

TIAN-TOM:
Quick update — PR #146 now has full invalid-input test coverage for the 25 model functions it touches (commit 9718049). Each model has parametrised tests with undefined, "string", null, NaN, and Infinity. 2012 tests passing.

Since the existing codebase already uses NaN returns for invalid/out-of-range inputs (check_standard_compliance_array, valid_range, utci inline checks, etc.), I think we should stick with that pattern rather than introducing throws. The numeric validation in #146 already covers the same 5 models plus the other 20, so there's quite a bit of overlap now.

yzq124537:
@FedericoTartarini 
Hi Federico, Tom and I have discussed the plan and decided that I will take issue #142 while Tom will handle the remaining issues. I've been working on it this week and pushed a new version — the PR description above has been updated.

FedericoTartarini:
@TIAN-TOM Could you please review the code and then tag me once your review is done and @yzq124537 has addressed your comments?

src/models/a_pmv.js
```diff
@@ -70,6 +70,19 @@ import { pmv } from "./pmv.js";
  * const result = a_pmv(tdb, tr, vr, rh, met, clo, a_coefficient, wme);
  * console.log(result) //output { a_pmv: 0.48 }
  */
+const A_PMV_SCHEMA = {
```
FedericoTartarini, line 73: I have a comment about this approach. Instead of having a schema for each model, should we consider having a general schema for all the models? What I'm trying to say here is that some of the variables, like the temperature, are shared across different models. So the same dictionary can be used across different functions, similar to what we are doing in PyThermal Comfort. If this approach is too complex we can keep the code as it is.

src/models/e_pmv.js
```diff
@@ -1,4 +1,8 @@
-import { round } from "../utilities/utilities.js";
+import { round, validateInputs } from "../utilities/utilities.js";
+
+const E_PMV_SCHEMA = {
+  e_coefficient: { type: "number" },
```
FedericoTartarini, line 4: This is not correct because you're not validating all the inputs I guess that you are validating them later on when you call the PMV, but we need to have a unified methodology. What do you think? I think this is particularly important because in the A_PMV model you are checking actually all the variables, not just the coefficients.

src/models/phs.js
```diff
@@ -103,6 +104,18 @@ const MET_WATT_PER_MET = 58.15;
...
+const PHS_SCHEMA = {
+  tdb: { type: "number" },
+  tr: { type: "number" },
+  v: { type: "number" },
+  rh: { type: "number" },
+  met: { type: "number" },
+  clo: { type: "number" },
+  wme: { type: "number" },
+  model: { enum: ["7933-2004", "7933-2023"] },
+  round: { type: "boolean", required: false },
```
FedericoTartarini, line 116: Don't we need to validate also the other inputs? Some of the inputs are passed as PHS KWargs

FedericoTartarini (changes requested):
@yzq124537 Thank you very much for working on this. I actually have provided some comments already. @TIAN-TOM Sorry for sending you two messages in such a short time but please review my comments as well and review the code whenever you have time.

TIAN-TOM (commented):
The schema-driven `validateInputs` + `assertNumber` approach fits the agreed TypeError/NaN split (TypeError for wrong types, NaN for out-of-range).

**On Federico's F1 (shared vs per-model schema):** I lean towards a shared definition for the common fields (`tdb`, `tr`, `v`, `rh`, `met`, `clo`, `units`, `limit_inputs`) with models composing them. The inconsistencies I flag below all come from duplicated field definitions getting out of sync. We can talk it through on Monday if you'd rather keep per-model.

---

### Three things that cross multiple files

**1. `units` lowercase handling is inconsistent across the library.**

| Behaviour | Models |
|---|---|
| `units.toUpperCase()` before validation → `"si"` accepted | `utci`, `pmv_ppd`, `a_pmv`, `cooling_effect`, `ankle_draft`, `adaptive_ashrae`, `adaptive_en` (7) |
| `units` passed raw → `"si"` throws | `clo_tout`, `vertical_tmp_grad_ppd`, `set_tmp`, `use_fans_heatwave`, `heat_index` (5, last via `options.units`) |

Same library, same parameter name, different behaviour. A shared field definition would fix this.

**2. `two_nodes.js` has the same schema gap as `phs` (F3).**

- `w_max` is accepted as a kwarg (`two_nodes.js:151`) and read at `:295`, but it's absent from `TWO_NODES_SCHEMA`
- `phs` and `two_nodes` are the only kwarg-accepting models in this PR with genuine gaps (`set_tmp` and `use_fans_heatwave` look clean on a quick read)

**3. `pmv.js` isn't in this PR.**

It sits between `e_pmv` and `pmv_ppd` in the call chain. Validation happens further down the call chain in `pmv_ppd`, not in `pmv` or `e_pmv` themselves. Either add a schema to `pmv` or document the delegation.

---

### A few smaller things

- **`required: false` applied inconsistently** — `limit_inputs`, `airspeed_control`, `units` are defaulted parameters in some schemas but not others. Harmless today because defaults prevent `undefined` reaching the validator, but the convention is unclear.
- **`solar_gain.js:123` mutates before validating** — `posture.toLowerCase()` runs before `validateInputs`, so a non-string posture surfaces a raw `TypeError: posture.toLowerCase is not a function` instead of the schema's enum message.
- **No direct validator tests** — `tests/utilities/utilities.test.js` wasn't updated. `assertNumber` and `validateInputs` only have indirect coverage through model tests.

---

Ping me when F1–F3 and the units thing are done and I'll take another pass.

src/models/e_pmv.js
```diff
@@ -1,4 +1,8 @@
-import { round } from "../utilities/utilities.js";
+import { round, validateInputs } from "../utilities/utilities.js";
+
+const E_PMV_SCHEMA = {
```
TIAN-TOM, line 3: Agreeing with Federico's comment on this line. Either expand `E_PMV_SCHEMA` to cover `tdb, tr, vr, rh, met, clo, e_coefficient, wme, units, limit_inputs` (mirroring A_PMV) so errors point to `e_pmv`, or leave the schema as-is and document in the JSDoc that `e_pmv` inherits validation from `pmv_ppd` via `pmv`. Without one of these, `e_pmv` has no validation of its own — invalid inputs are only caught further down in `pmv_ppd`.

Also: `E_PMV_SCHEMA` is declared at line 3, wedged between two `import` statements. Every other model in this PR puts the schema immediately before `export function` (after JSDoc). Move it below the JSDoc for consistency.

src/models/phs.js
TIAN-TOM, line 116: +1 on Federico's comment. `PHS_SCHEMA` covers 9 params but `defaults_kwargs` (line 131) declares 18 keys. None of `i_mst, a_p, drink, weight, height, walk_sp, theta, acclimatized, duration, f_r, t_sk, t_cr, t_re, t_cr_eq, t_sk_t_cr_wg, sweat_rate_watt, evap_load_wm2_min` are validated.

Separate concern on the same function: `posture` is declared in JSDoc as `number | "sitting" | "standing" | "crouching"` but isn't in the schema. The if/else at line 156 handles the three known strings, but there's no else branch. A caller passing `"invalid"` or `5` silently falls through to `_radiating_area_dubois` (line 241), which returns `0.67` (the crouching branch) via the `?:` chain. Either add an enum to the schema (1|2|3 + the three strings) or throw in the else branch.

src/models/utci.js
```diff
@@ -69,6 +84,19 @@ export function utci(
   return_stress_category = false,
   limit_inputs = true,
 ) {
+  validateInputs(
+    {
+      tdb,
+      tr,
+      v,
+      rh,
+      units: units.toUpperCase(),
```
TIAN-TOM, line 93: This line passes `units.toUpperCase()` to the validator, so `"si"` / `"ip"` are accepted and the rest of the function uses `units.toLowerCase() == "ip"` (line 102), which works. That's fine on its own.

The issue is that this normalisation isn't applied consistently across the library. `clo_tout.js:39`, `vertical_tmp_grad_ppd.js:73`, `set_tmp.js`, `use_fans_heatwave.js`, and `heat_index.js` pass `units` raw (the last via `options.units`), so lowercase throws there. Meanwhile `pmv_ppd`, `a_pmv`, `cooling_effect`, `ankle_draft`, `adaptive_ashrae`, `adaptive_en` all normalise like utci does.

Whichever behaviour we standardise on (always case-insensitive, or always strict `"SI"` / `"IP"`), it should be the same across all 12 models.

src/models/two_nodes.js
```diff
@@ -113,6 +113,23 @@ import { round } from "../utilities/utilities.js";
 }
  *
  */
+const TWO_NODES_SCHEMA = {
```
TIAN-TOM, line 116: `w_max` is declared in `defaults_kwargs` at line 151 and read at line 295, but isn't in `TWO_NODES_SCHEMA`. Same gap pattern as Federico's F3 against `phs`. A caller passing `{ w_max: "not a number" }` bypasses the schema check — any truthy non-numeric value silently skips the formula block in `calculate_w_max` and returns the initialised `0`, rather than surfacing the bad input.

Scanned the other kwarg-accepting models — `set_tmp` and `use_fans_heatwave` look clean. `phs` (F3) and `two_nodes` are the only genuine gaps in this PR.

JitaoFu: I'd like to add it, but it requires a small change to calculate_w_max to align with the validation logic. Let me know if this overlaps with what you're working on.

### The issue with the current implementation:
The user can never actually pass a numeric `w_max` value to override the calculation in `calculate_w_max`. When a number like 0.6 is passed, !0.6 is false, so the if block is skipped and wMax returns 0 instead of 0.6. This is a bug.

``` javascript
function calculate_w_max(airSpeed, clo, kwargs) {
  let wMax = 0;
  if (!kwargs.w_max) {
    wMax = 0.38 * Math.pow(airSpeed, -0.29);
    if (clo > 0) {
      wMax = 0.59 * Math.pow(airSpeed, -0.08);
    }
  }
  return wMax;
}
```


### Proposed fix:
- Remove w_max: false from defaults_kwargs so it stays undefined when not provided.
``` javascript
const defaults_kwargs = {
  calculate_ce: false,
  round: true,
  max_sweating: 500,
};
```
- Fix calculate_w_max:
``` javascript
function calculate_w_max(airSpeed, clo, kwargs) {
  if (kwargs.w_max) return kwargs.w_max;
  if (clo > 0) return 0.59 * Math.pow(airSpeed, -0.08);
  return 0.38 * Math.pow(airSpeed, -0.29);
}
```
- Add validation logic for w_max: w_max: { type: "number", required: false }

### Result:
If w_max is not provided → undefined → validation skipped, auto-calculation runs as before. 
If w_max is provided → must be a number, otherwise TypeError, and the value is used directly in calculate_w_max.

Is this change within your current scope? If not, I can take it.

src/models/a_pmv.js
JitaoFu: I've looked into this further, and here's what I found.

#### Python's Native Dataclass Support Makes Shared Schema Viable, But JavaScript Port Adds More Complexity Than It Saves
- The Python dataclass mechanism in BaseInputs natively handles the case where different models pass different subsets of parameters. Fields that are None are automatically skipped during validation, which means Python can do it without any extra code. 
- To achieve the same effect in JavaScript, we would need to introduce additional validation logic and restructure the existing validateInputs function, which end up more or less the same with the current code structure. In other words, While this would reduce some repeated field definitions across models, it would simultaneously introduce greater complexity, and reduce both readability and maintainability. 
- Therefore, although the shared schema approach works naturally in Python due to language-level support, replicating it in JavaScript may results in a net increase in complexity rather than a simplification.

#### Advantages of the existing input validation design
- The current per-model schema design does involve some repetition, but it is straightforward to read — it is immediately clear which parameters are being validated for each model. 
- It is also easy to modify, since changes can be made directly in the relevant model file without affecting others.

On balance, I'd lean towards keeping the existing per-model schema design, but happy to follow your preference.

src/models/e_pmv.js
JitaoFu: Got it, I will add the same checking logic to e_pmv, pmv, pmv_ppd_ashrae, pmv_ppd_iso model to ensure a unified methodology.

src/models/two_nodes.js
TIAN-TOM: Yeah, go ahead, this belongs in #145.

I also touched `calculate_w_max` in #154, but that PR is really about the test harness, so I dropped that overlap from #154. #145 is the better place for this shared fix.

One thing to fix before this lands: the shared validation data still uses `w_max: false` as the auto-calculation sentinel. `ts_two_nodes_gagge.json` row 5 (0-indexed) has that value, matching pythermalcomfort's default `False` sentinel. With the current number schema, that row throws before it reaches `calculate_w_max`.

I think the smallest fix is to normalise before `validateInputs`:

```js
if (joint_kwargs.w_max === false) joint_kwargs.w_max = undefined;
```

Then the new guard still takes the auto-calculation path, while `true`, strings, and `NaN` still fail validation.

JitaoFu: @TIAN-TOM Nice catch on the w_max: false Python sentinel, that was a subtle one hiding in the test data.                                                                                                        
All changes are done, Could you do another review when you get a chance? Thanks.

src/models/phs.js
JitaoFu: Thanks for pointing that out. I've completed the type checking logic for all inputs (including Kwargs) in the new commit. All changes are addressed, Could you do another review when you get a chance? Thanks.

src/models/two_nodes.js
TIAN-TOM: Verified in 62c6d5b: `w_max` is in the schema, `false` is normalised to `undefined` before validation, and `calculate_w_max` has been rewritten. CI is green across all five checks. Resolving this thread.

src/models/phs.js
TIAN-TOM: The expanded `PHS_SCHEMA` in e12375f covers every tunable kwarg and adds `posture: { enum: [1, 2, 3] }`, so invalid posture values now fail validation instead of falling through to the crouching branch. Both parts of my comment are addressed.

src/models/utci.js
TIAN-TOM: 72bf31e adds uppercase units normalisation in `clo_tout`, `vertical_tmp_grad_ppd`, `set_tmp`, `use_fans_heatwave`, and `heat_index` via `options.units`. That addresses the inconsistency I flagged.

TIAN-TOM:
@FedericoTartarini Re-review on 62c6d5b is done. The PHS kwargs coverage gap is closed by e12375f, and my three follow-up threads are resolved from my side: `two_nodes` `w_max`, posture enum, and units normalisation. The remaining non-outdated open thread is the shared-vs-per-model schema question on `a_pmv.js`, where Jitao leans towards keeping per-model schemas and left the final call to you.

FedericoTartarini:
okay let's keep it per-schema

PR #146, opened 2026-04-11
refactor: add input type validation to core model functions

TIAN-TOM:
Adds numeric input validation to 25 model functions (out of 28 total exports). Each function now checks that numeric parameters are finite numbers before calculating. Invalid inputs (undefined, null, strings, NaN, Infinity) trigger an early return with NaN in the same shape the function normally uses — most return an object like { pmv: NaN, ppd: NaN }, but pmv and wbgt (round=true) return bare NaN directly. The three not touched are JOS3 (class, out of scope), pmv_ppd_ashrae and pmv_ppd_iso (thin wrappers — their inputs get validated when they call pmv_ppd internally).

Ref #142. This covers type validation only. I left range validation (limit_inputs boundaries like 10 ≤ tdb < 40 for PMV per ASHRAE 55) for a follow-up PR — need to map out the pythermalcomfort ranges properly first.

Implementation: a small utility src/utilities/validate_inputs.js with isValidNumber (typeof === 'number' && isFinite) and allValidNumbers. Each model calls the guard at the top. No calculation logic changed.

Also fixed two things I noticed along the way:
- utci mapping() was returning a stress category string for NaN input; added an isFinite guard so it returns NaN instead
- adaptive_en JSDoc example showed true for acceptability when t_running_mean = 9 (out of range), corrected to false

Both helpers are tagged @ignore to keep them out of the generated docs.

Every model has parametrised invalid-input tests covering each numeric parameter with all five bad-value types. Models with extra validated params (clo, v_ankle, a_coefficient, etc.) have dedicated rows. wbgt has separate sub-blocks for round=true and round=false. 68 suites, 2012 tests, 0 failures.

This PR does not add checks for most boolean/enum parameters (units, limit_inputs, round, standard). phs posture is a special case — it accepts both numbers and strings, so it has a dedicated type check alongside the numeric guard.

### DIFF
--- a/src/models/a_pmv.js
+++ b/src/models/a_pmv.js
@@ -1,4 +1,5 @@
 import { round } from "../utilities/utilities.js";
+import { allValidNumbers } from "../utilities/validate_inputs.js";
 import { pmv } from "./pmv.js";
 
 /**
@@ -81,6 +82,10 @@ export function a_pmv(
   wme = 0,
   kwargs = {},
 ) {
+  if (!allValidNumbers(tdb, tr, vr, rh, met, clo, a_coefficient, wme)) {
+    return { a_pmv: NaN };
+  }
+
   const default_kwargs = {
     units: "SI",
     limit_inputs: true,
--- a/src/models/adaptive_ashrae.js
+++ b/src/models/adaptive_ashrae.js
@@ -4,6 +4,7 @@ import {
   round,
   units_converter,
 } from "../utilities/utilities.js";
+import { allValidNumbers } from "../utilities/validate_inputs.js";
 import { get_ce } from "./adaptive_en.js";
 
 /**
@@ -88,6 +89,18 @@ export function adaptive_ashrae(
   units = "SI",
   limit_inputs = true,
 ) {
+  if (!allValidNumbers(tdb, tr, t_running_mean, v)) {
+    return {
+      tmp_cmf: NaN,
+      tmp_cmf_80_low: NaN,
+      tmp_cmf_80_up: NaN,
+      tmp_cmf_90_low: NaN,
+      tmp_cmf_90_up: NaN,
+      acceptability_80: false,
+      acceptability_90: false,
+    };
+  }
+
   const standard = "ASHRAE";
   if (units.toUpperCase() === "IP") {
     ({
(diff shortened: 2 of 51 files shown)


yzq124537:
I'd like to share some thoughts on the validation approach. I think input validation and range checking serve different purposes and should be handled separately.

- Input validation is a front-line guard that checks whether the inputs are of the correct type. When invalid types are passed in, the function should fail fast and throw a TypeError with a clear message indicating which parameter is invalid. This makes it immediately obvious to the caller what went wrong and where.

- Range checking, on the other hand, is business logic that determines whether inputs fall within the standard's applicability limits. Returning NaN here makes sense, it is also consistent with the Python reference implementation.

- The approach I implemented was inspired by the Python pythermalcomfort reference implementation, which raises TypeError for invalid input types through its BaseInputs validation framework.

Combining both into a single NaN return makes it harder to distinguish between a type error and a range violation. It also loses the specific parameter name from the error message, which makes debugging more difficult.
I'd be happy to discuss this further — happy to hear your thoughts.

TIAN-TOM:
I think you're right that type mismatch and range validation are different cases, and pythermalcomfort does separate them (`TypeError` via `validate_type`, NaN via `limit_inputs`).

I kept NaN here because the JS side already uses NaN for invalid inputs in several places — `check_standard_compliance_array`, `valid_range`, and the `utci` inline checks all return NaN rather than throwing. Adding TypeError for type checks would mean two different error-handling conventions in the same library.

If we want to distinguish type errors from out-of-range values, that's probably a broader API decision rather than a one-off change in this PR. Happy to get Federico's view on it at tomorrow's meeting.

yzq124537:
That's a fair point about consistency. 

I think the disagreement comes down to how we define 'input validation' — `check_standard_compliance_array` and `valid_range` handle out-of-range values, which I'd classify as **range checks** rather than **input validation**. 

In my view they're doing different things, so having separate error-handling conventions for each would still be consistent in intent. 

But I agree this is a broader API decision and maybe we can get Federico's view at tomorrow's meeting.

TIAN-TOM:
Closing in favour of #145, which now covers the same 25 models using the TypeError approach agreed at the 2026-04-13 team meeting.

PR #149, opened 2026-04-16
ci: automate GitHub Release creation alongside NPM publish

Pyqren:
Currently, our GitHub Actions workflow automatically publishes a new package to the NPM registry whenever a new v* tag is pushed. However, this does not automatically create a Release on our GitHub repository. This has led to our GitHub Releases page becoming outdated which can confuse users looking for the latest version.

---

### What was changed:

- Added explicit `contents: write` permissions and implemented the `softprops/action-gh-release` step in `.github/workflows/publish.yml`.
- Updated docs_theme/contributing.md to reflect the new release workflow.

Pushing a new v* tag will now automatically trigger both an NPM publish and the creation of a new GitHub Release. The release will automatically generate its own release notes (changelog) based on the PRs merged since the last tag.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -53,6 +53,6 @@ jobs:
       - run: npm publish --provenance
 
       - name: Create GitHub Release
-        uses: softprops/action-gh-release@v2
+        uses: softprops/action-gh-release@v3
         with:
           generate_release_notes: true
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -38,7 +38,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: checks
     permissions:
-      contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -51,3 +51,8 @@ jobs:
 
       - run: npm ci
       - run: npm publish --provenance
+
+      - name: Create GitHub Release
+        uses: softprops/action-gh-release@v2
```
FedericoTartarini: update this

FedericoTartarini (changes requested):

Pyqren:
@FedericoTartarini Updated to v3 following our feedback. I opted for a specific version instead of latest version as it seems like the more stable option.

FedericoTartarini (approved):

Issue #153, opened 2026-04-19
two_nodes: nSimulation initial value off-by-one vs pythermalcomfort (Row 7)


TIAN-TOM:
Row 7 of `ts_two_nodes_gagge.json` fails on 5 fields once the `two_nodes` harness is repaired, with no caller overrides involved.

Inputs: `tdb=45, tr=45, v=1.1, rh=20, met=3, clo=0.2`.

At [`src/models/two_nodes.js:497`](https://github.com/FedericoTartarini/jsthermalcomfort/blob/1714cfa/src/models/two_nodes.js#L497), JS starts with `let nSimulation = 0;`. pythermalcomfort starts the equivalent counter at `1`, which gives one more convergence iteration.

Changing this to `let nSimulation = 1;` brings the five failing Row 7 fields (`e_rsw`, `e_skin`, `m_rsw`, `pmv_set`, `disc`) within tolerance against the validation dataset / pythermalcomfort 3.9.1. For example: `e_rsw` `219.31 -> 244.25`, `e_skin` `229.69 -> 253.12`, `m_rsw` `322.51 -> 359.19`, `pmv_set` `2.61 -> 2.24`, `disc` `4.70 -> 5.34`.

This change targets Row 7's failure mode. Row 6's passing fields remain within tolerance with the fix applied; the Row 6 `disc` divergence stems from a separate formula issue and is unaffected by this counter.

Related: #152 (independent `disc`-formula divergence on Row 6).

Issue #152, opened 2026-04-19
two_nodes: disc field diverges substantially on Row 6 (expected 6.0, actual ~16.56)


TIAN-TOM:
Row 6 of `ts_two_nodes_gagge.json` fails on the `disc` field. Inputs: `tdb=40, tr=40, v=1.1, rh=50, met=2, clo=0.5, w_max=0.2`.

| Source | `disc` value |
| --- | --- |
| `ts_two_nodes_gagge.json` (reference) | 6.0 |
| pythermalcomfort 3.9.1 `two_nodes_gagge` | 6.0 |
| jsthermalcomfort | ~16.56 |

The JSON reference agrees with pythermalcomfort 3.9.1, so the divergence is on the JS side. The delta is ~10.56, far outside the per-field tolerance of 0.1.

Context: before the separate `w_max` kwarg override bug in `calculate_w_max()` was fixed, Row 6's `disc` delta was ~2.9 — closer to the reference, but only because `w` was incorrectly 0 instead of 0.2, and the disc calculation against that wrong `w` happened to sit nearer to 6.0. Once `w` is correct, the underlying JS-side `disc`-formula bug becomes visible (delta ~10.6).

The six other Row 6 fields (`t_core`, `e_skin`, `m_rsw`, `pmv_set`, `t_skin`, `w`) all pass, so the remaining issue is isolated to `disc`.

Related: #153 (independent `nSimulation` off-by-one affecting Row 7).

The `w_max` kwarg override bug referenced above was fixed in #154.

Issue #135, opened 2025-04-13
Some Models Failing Tests in the Latest Build


rcqz:
## **Description**  
When running the test suite (via `npm test`), several of the models in the project are failing some or all of their tests.

## **Affected Models**

- `pet_steady.js`
    
- `phs.js`
    
- `two_nodes.js`
    

## **Expected Behaviour**  
Given the validation tables, these model should produce outputs that match the validation tables.
   
**Example**: When running ` npm run test -- tests/models/pet_steady.test.js `, two of the test cases fail. The following shows the output from the failed test cases. 

```
 FAIL  tests/models/pet_steady.test.js (8.273 s)
  pet_steady
    ✓ Test case #0 (7 ms)
    ✓ Test case #1 (1 ms)
    ✓ Test case #2
    ✓ Test case #3
    ✓ Test case #4 (1 ms)
    ✓ Test case #5
    ✓ Test case #6 (1 ms)
    ✕ Test case #7 (68 ms)
    ✕ Test case #8 (3 ms)

  ● pet_steady › Test case #7

    expect(received).toBeCloseTo(expected, precision)

    Expected: 44.63
    Received: 43.05

    Expected precision:    0.1
    Expected difference: < 0.4
    Received difference:   1.5800000000000054

      81 |           expect(actualValue).toBeNaN();
      82 |         } else {
    > 83 |           expect(actualValue).toBeCloseTo(expectedValue, tol);
         |                               ^
      84 |         }
      85 |       } else {
      86 |         // For booleans or other types

      at tests/models/testUtils.js:83:31
          at Array.forEach (<anonymous>)
      at validateResult (tests/models/testUtils.js:52:34)
      at tests/models/pet_steady.test.js:20:5

  ● pet_steady › Test case #8

    expect(received).toBeCloseTo(expected, precision)

    Expected: 32.21
    Received: 31.69

    Expected precision:    0.1
    Expected difference: < 0.4
    Received difference:   0.5199999999999996

      81 |           expect(actualValue).toBeNaN();
      82 |         } else {
    > 83 |           expect(actualValue).toBeCloseTo(expectedValue, tol);
         |                               ^
      84 |         }
      85 |       } else {
      86 |         // For booleans or other types

      at tests/models/testUtils.js:83:31
          at Array.forEach (<anonymous>)
      at validateResult (tests/models/testUtils.js:52:34)
      at tests/models/pet_steady.test.js:20:5

[...]

PR #154, opened 2026-04-19
test(two_nodes): repair validation harness and expose row failures

TIAN-TOM:
Partial fix for #135.

Note: force-pushed this PR to drop commit `1714cfa5` (the `calculate_w_max` source fix). This branch now only changes `tests/models/two_nodes.test.js`; #145 is carrying the source fix.

The `two_nodes` test harness wasn't exercising the validation dataset. The main issue: the test destructured `{ inputs, expected }`, but `ts_two_nodes_gagge.json` stores results under `outputs`, so `expected` was always `undefined` and the per-row skip guard silently dropped all 10 dataset rows. The only real assertions running were the two meta-tests in the second `describe`.

Three more harness issues were hiding behind that (rows are 0-indexed to match the JSON array):

1. Flat override fields (`w_max`, `max_sweating`) sit at the top of `inputs` in the JSON, not nested under `inputs.kwargs`, so overrides in rows 6 and 8 were dropped by the harness. Row 5's `w_max: false` is the default-formula sentinel, so dropping it had no behavioural effect.
2. The harness didn't force `round: false`, so rows with 2-decimal reference values, like row 4's `w=0.17`, failed on a rounded `w=0.2`. `round: false` also switches output keys from snake_case to camelCase, so the harness now normalises the keys.
3. All rows ran inside a single `it(... forEach ...)`, which stopped at the first failure. This now uses `test.each` so every row reports independently.

Two plumbing fixes also rolled in: `loadTestData` was called with `"twoNodes"` as the second argument where the signature expects a boolean `returnArray`, and the return was destructured as `tolerance` where the helper returns `tolerances`.

Current result on this branch: the harness now runs all 10 dataset rows. `two_nodes.test.js` registers 12 Jest tests (10 rows + 2 meta-tests): 10 pass, 2 fail.

- Row 6 fails on `t_core`, `e_skin`, `m_rsw`, `t_skin`, `w`, `disc`, and `pmv_set`. The root cause is the `calculate_w_max` guard zeroing `w_max` on this branch, which cascades through the evaporative-sweat path (`w=0`, `m_rsw=0`, and so on). #145 should unblock Row 6; the residual `disc` mismatch is tracked in #152 but can only be isolated once the w_max cascade is gone.
- Row 7 fails on an `nSimulation` initial-value off-by-one, tracked in #153.

So this PR is test-harness only: it exposes the real validation failures but does not fix the model bugs. CI will stay partially red until #145, #152, and #153 land.

### DIFF
--- a/tests/models/two_nodes.test.js
+++ b/tests/models/two_nodes.test.js
@@ -1,73 +1,89 @@
-import { expect, describe, it, beforeAll } from "@jest/globals";
-import { loadTestData } from "./testUtils";
+import { expect, describe, test, it } from "@jest/globals";
+import { loadTestData, validateResult } from "./testUtils";
 import { two_nodes } from "../../src/models/two_nodes";
 import { testDataUrls } from "./comftest"; // Import all test URLs from comftest.js
 
 // Use the URL from comftest.js to fetch data for two_nodes tests
 const testDataUrl = testDataUrls.twoNodes; // Ensure the correct URL is added in comftest.js
 
-let testData;
-let tolerance;
+// Load data at module scope so test.each can register one test per row.
+const { testData, tolerances } = await loadTestData(testDataUrl, false);
 
-// Load data before tests
-beforeAll(async () => {
-  const result = await loadTestData(testDataUrl, "twoNodes"); // Load data from URL using loadTestData
-  testData = result.testData;
-  tolerance = result.tolerance || 0.1; // Set a default tolerance
-});
-
-// General test function to verify results for different test functions
-function runTest(testFunction, inputs, expected) {
-  const {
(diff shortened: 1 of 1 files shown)


TIAN-TOM:
CI red is expected here. This PR fixes the harness (which was silently skipping all 10 validation rows) and the Row 6 w_max override bug. The two remaining failures are the Row 6 disc-formula divergence (tracked in #152) and Row 7 nSimulation off-by-one (tracked in #153), both surfaced by the harness repair and scoped to separate issues.

TIAN-TOM:
Closing in favour of #162, which consolidates this PR with the disc-formula fix in a single review.

PR #156, opened 2026-04-20
feat: add named exports

Pyqren:
Added named exports to various index.js files in order to enable compact importing of the functions to comfort-tool.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,3 +1,8 @@
+export * from "./utilities/index.js";
+export * from "./psychrometrics/index.js";
+export * from "./models/index.js";
+export * from "./supa.js";
+
 import utilities from "./utilities/index.js";
 import psychrometrics from "./psychrometrics/index.js";
 import models from "./models/index.js";
--- a/src/index.js
+++ b/src/index.js
@@ -1,7 +1,6 @@
 export * from "./utilities/index.js";
 export * from "./psychrometrics/index.js";
 export * from "./models/index.js";
-export * from "./supa.js";
 
 import utilities from "./utilities/index.js";
 import psychrometrics from "./psychrometrics/index.js";
--- a/src/models/index.js
+++ b/src/models/index.js
@@ -60,7 +60,6 @@ export {
   vertical_tmp_grad_ppd,
   wbgt,
   wc,
-  JOS3,
 };
 
 export default {
--- a/src/models/index.js
+++ b/src/models/index.js
@@ -32,6 +32,37 @@ import { JOS3 } from "./JOS3.js";
  * @name models
  * @docname Comfort Models
  */
+export {
+  adaptive_ashrae,
+  adaptive_en,
+  a_pmv,
+  athb,
+  e_pmv,
+  at,
+  ankle_draft,
+  clo_tout,
+  cooling_effect,
+  discomfort_index,
+  heat_index,
+  humidex,
+  net,
+  phs,
+  pet_steady,
+  pmv_ppd,
+  pmv_ppd_ashrae,
+  pmv_ppd_iso,
+  pmv,
+  solar_gain,
+  set_tmp,
+  two_nodes,
+  utci,
+  use_fans_heatwaves,
+  vertical_tmp_grad_ppd,
+  wbgt,
+  wc,
+  JOS3,
+};
+
 export default {
   adaptive_ashrae,
   adaptive_en,
--- a/src/psychrometrics/index.js
+++ b/src/psychrometrics/index.js
@@ -12,6 +12,8 @@ import { psy_ta_rh } from "./psy_ta_rh.js";
  * @name psychrometrics
  * @docname Psychrometrics
  */
+export { p_sat, p_sat_torr, psy_ta_rh, t_o, enthalpy, t_wb, t_dp, t_mrt };
+
 export default {
   p_sat,
   p_sat_torr,
--- a/src/utilities/index.js
+++ b/src/utilities/index.js
@@ -8,11 +8,7 @@ import {
   met_typical_tasks,
   clo_typical_ensembles,
   clo_individual_garments,
-  check_standard_compliance,
   check_standard_compliance_array,
-  round,
-  transpose_sharp_altitude,
-  valid_range,
 } from "./utilities.js";
 
 /**
@@ -30,11 +26,7 @@ export {
   met_typical_tasks,
   clo_typical_ensembles,
   clo_individual_garments,
-  check_standard_compliance,
   check_standard_compliance_array,
-  round,
-  transpose_sharp_altitude,
-  valid_range,
 };
 
 export default {
@@ -47,11 +39,7 @@ export default {
   met_typical_tasks,
   clo_typical_ensembles,
   clo_individual_garments,
-  check_standard_compliance,
   check_standard_compliance_array,
-  round,
-  transpose_sharp_altitude,
-  valid_range,
 };
 
 /**
--- a/src/utilities/index.js
+++ b/src/utilities/index.js
@@ -39,7 +39,6 @@ export default {
   met_typical_tasks,
   clo_typical_ensembles,
   clo_individual_garments,
-  check_standard_compliance_array,
 };
 
 /**
--- a/src/utilities/index.js
+++ b/src/utilities/index.js
@@ -8,6 +8,11 @@ import {
   met_typical_tasks,
   clo_typical_ensembles,
   clo_individual_garments,
+  check_standard_compliance,
+  check_standard_compliance_array,
+  round,
+  transpose_sharp_altitude,
+  valid_range,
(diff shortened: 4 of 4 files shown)


src/models/index.js
```diff
@@ -32,6 +32,37 @@ import { JOS3 } from "./JOS3.js";
...
+  pmv,
+  solar_gain,
+  set_tmp,
+  two_nodes,
+  utci,
+  use_fans_heatwaves,
+  vertical_tmp_grad_ppd,
+  wbgt,
+  wc,
+  JOS3,
```
FedericoTartarini: Have we fully tested the Jaws 3 model? If not, we should potentially remove it from this package because we only want to include models that have been tested and we are 100% sure that they are accurate.

src/psychrometrics/index.js
```diff
@@ -12,6 +12,8 @@ import { psy_ta_rh } from "./psy_ta_rh.js";
  * @name psychrometrics
  * @docname Psychrometrics
  */
+export { p_sat, p_sat_torr, psy_ta_rh, t_o, enthalpy, t_wb, t_dp, t_mrt };
```
FedericoTartarini, line 15: Perfect that you are exporting these functions and you should use them in the comfort tool.

src/utilities/index.js
```diff
@@ -8,13 +8,35 @@ import {
...
+  body_surface_area,
+  v_relative,
+  clo_dynamic,
+  units_converter,
+  running_mean_outdoor_temperature,
+  f_svv,
+  met_typical_tasks,
+  clo_typical_ensembles,
+  clo_individual_garments,
+  check_standard_compliance,
```
FedericoTartarini: The function like the checkStandardCompliance should not be exported because this is a utility function that is used by the package internally and users of jsthermalcomofort should not be using it. We should hide this function.

src/utilities/index.js
```diff
@@ -25,6 +47,11 @@ export default {
   met_typical_tasks,
   clo_typical_ensembles,
   clo_individual_garments,
+  check_standard_compliance,
```
FedericoTartarini: Why are we exporting the same function three times in this file? See also my previous comment that we should not export this function at all.

src/index.js
```diff
@@ -1,3 +1,8 @@
+export * from "./utilities/index.js";
+export * from "./psychrometrics/index.js";
+export * from "./models/index.js";
+export * from "./supa.js";
```
FedericoTartarini: What is supa.js?

FedericoTartarini (changes requested):
Could you please address my comments and then tag me again?

src/psychrometrics/index.js
Pyqren: Will do.

src/models/index.js
Pyqren: Good call. I'll remove it from named exports for now and leave it for internal use, as the team's still working on it.

src/index.js
Pyqren: This is a file in `srs/` that has array helpers for the JOS3 model. Since we're removing JOS3 from named exports, `supa.js` is no longer needed. I'll remove it as well.

src/utilities/index.js
Pyqren: I'll remove  `round`, `valid_range`, and `transpose_sharp_altitude` from exports so they no longer show up as options for user. I'll remove `check_standard_compliance` and keep `check_standard_compliance_array` as we use the array version for comfort-tool.

src/utilities/index.js
Pyqren: Fixed.

Pyqren:
> Could you please address my comments and then tag me again?

@FedericoTartarini All comments addressed, files updated.

src/utilities/index.js
```diff
@@ -8,13 +8,27 @@ import {
...
+  body_surface_area,
+  v_relative,
+  clo_dynamic,
+  units_converter,
+  running_mean_outdoor_temperature,
+  f_svv,
+  met_typical_tasks,
+  clo_typical_ensembles,
+  clo_individual_garments,
+  check_standard_compliance_array,
```
FedericoTartarini, line 29: Why do we still have this function despite we have removed all the array versions of the function?

src/utilities/index.js
```diff
@@ -8,13 +8,27 @@ import {
   met_typical_tasks,
   clo_typical_ensembles,
   clo_individual_garments,
+  check_standard_compliance_array,
```
FedericoTartarini, line 11: Same comment as below.

src/utilities/index.js
```diff
@@ -25,6 +39,7 @@ export default {
...
```
FedericoTartarini: Same comment as before.

FedericoTartarini (changes requested):
@Pyqren Thank you so much for working on this. However, I didn't see your response to my previous question and I'm not understanding why we are exporting the same function multiple times.

I also really appreciate your work and you are one of the most active students which I'm really grateful about it. However, it would be fantastic if we could just be a bit more careful about the details, like for instance including a function that shouldn't be there in the first place. Thank you very much for your work.

I hope you understand that these are libraries that many people may be using in the community so we need to be 100% sure that they are working well and well maintained.

src/utilities/index.js
Pyqren: I removed `check_standard_compliance` and kept `check_standard_compliance_array` as the current implementation is using the array version for comfort-tool. Should we not be using the array version?

Pyqren:
> @Pyqren Thank you so much for working on this. However, I didn't see your response to my previous question and I'm not understanding why we are exporting the same function multiple times.
> 
> I also really appreciate your work and you are one of the most active students which I'm really grateful about it. However, it would be fantastic if we could just be a bit more careful about the details, like for instance including a function that shouldn't be there in the first place. Thank you very much for your work.
> 
> I hope you understand that these are libraries that many people may be using in the community so we need to be 100% sure that they are working well and well maintained.

@FedericoTartarini No problem at all, I appreciate your thorough feedback to ensure our code is as good as it could be.

I have unresolved the comments, hopefully they are accessible again.

src/utilities/index.js
Pyqren: In here, we're bringing the `check_standard_compliance_array` function from the `utilities.js` file into this `index.js` file's local scope so it can be exported. This makes it so that the exports are centralized.

src/utilities/index.js
Pyqren: See response [here](https://github.com/FedericoTartarini/jsthermalcomfort/pull/156#discussion_r3120710890) for context. I believe this one may not be necessary if it's already present in the above export, however, and can be removed. If you approve using the array version, I'll remove it from here and keep the other one.

src/utilities/index.js
FedericoTartarini: I am honestly a little bit confused. I recall that we have decided to remove all the array versions of the functions. Therefore we don't need the checkStandardComplianceArray function. @TIAN-TOM could you also please review this?

src/utilities/index.js
FedericoTartarini: Okay I see the point

src/utilities/index.js
Pyqren: @TIAN-TOM @FedericoTartarini 
Currently, `check_standard_compliance_array` contains a unique validation logic for ASHRAE airspeed control limits that the `check_standard_compliance` function does not have. At the moment, the comfort-tool dashboard relies on this specific logic to determine if the status is Compliant or not.

If you'd like, I can switch it back to the scalar one and refactor the code to take scalar values, but we may lose accuracy. I recommend we refactor the scalar function to include this logic in jsthermalcomfort first before we switch in comfort-tool.

Please let me know your thoughts.

TIAN-TOM: @FedericoTartarini I checked the implementation, and the code supports Pyqren's point: `check_standard_compliance_array` contains ASHRAE airspeed-control validation that `check_standard_compliance` does not currently implement. The comfort-tool PMV path uses that helper when deriving the compliance result shown as "Compliant" / "Out of range".

For the earlier `*_array` cleanup: PR #141 removed three utilities entries from `utilities/index.js` (`v_relative_array`, `clo_dynamic_array`, `units_converter_array`) and their function bodies. Before #156, `check_standard_compliance_array` was not in `utilities/index.js`; it existed as a named export from `utilities/utilities.js`. PR #156 is now adding it to the new named-export block, which is a different export path from the default-export object touched in #141.

My recommendation for this PR is to keep `check_standard_compliance_array` in the named exports for now. Moving the airspeed-control logic into the scalar helper first, then retiring the `_array` export, is the cleaner end state, but that is separate from the named-exports scope and should not block #156.

FedericoTartarini (approved):

PR #157, opened 2026-04-20
fix: classify UTCI stress categories by range, not nearest boundary

### LINKED ISSUE #147: Bug: UTCI mapping() uses nearest-neighbour lookup instead of threshold-based classification
## Problem

The `mapping()` function in `src/models/utci.js` (lines 127-153 on main) classifies UTCI values using a nearest-neighbour binary search over the boundary array `[-40, -27, -13, 0, 9, 26, 32, 38, 46, 1000]`. It picks whichever boundary value is numerically closest to the input, rather than checking which interval the value falls into.

This shifts the effective category transitions to the midpoint between adjacent boundaries. For example, the transition from "no thermal stress" to "moderate heat stress" should happen at 26 but actually happens at 29 (the midpoint of 26 and 32). The transition from "very strong heat stress" to "extreme heat stress" should be at 46 but the midpoint with the sentinel value 1000 is 523, so "extreme heat stress" is never returned in any realistic range.

Numerical UTCI values are not affected. Only the `stress_category` string returned when `return_stress_category = true` is wrong.

```js
import { utci } from './src/models/utci.js';

// UTCI = 15.2 falls in the "no thermal stress" range (9 < UTCI <= 26),
// but mapping() returns "slight cold stress" because 15.2 is closer to 9 than to 26
const result = utci(15, 15, 0.5, 50, 'SI', true);
console.log(result.stress_category);
// actual on main: "slight cold stress"
// expected per standard: "no thermal stress"
```

About 41 of 101 integer values in the -50 to +50 range disagree with pythermalcomfort's classification.

The comfort-tool frontend (FedericoTartarini/comfort-tool#2) passes the library's `stress_category` to the UI. The chart background bands use correct local boundaries, but the text label comes from `mapping()`, so they can contradict each other.

## Suggested fix

Replace the binary search with a threshold chain matching pythermalcomfort's `np.digitize(value, bins, right=True)` convention, where each boundary value belongs to the lower category:

```js
function mapping(val) {
  if (!isFinite(val)) return NaN;
  if (val <= -40) return stress_categories[0];
  if (val <= -27) return stress_categories[1];
  if (val <= -13) return stress_categories[2];
  if (val <= 0)   return stress_categories[3];
  if (val <= 9)   return stress_categories[4];
  if (val <= 26)  return stress_categories[5];
  if (val <= 32)  return stress_categories[6];
  if (val <= 38)  return stress_categories[7];
  if (val <= 46)  return stress_categories[8];
  return stress_categories[9];
}
```

PR #146 added an `!isFinite` guard to `mapping()` so `mapping(NaN)` returns `NaN` instead of a category string, but the core boundary bug is still there. The `stress_categories` and `stress_categories_vals` arrays themselves are correct.

TIAN-TOM:
## What this PR does

- Replace `mapping()` in `src/models/utci.js` with a `<=` threshold chain matching `np.digitize(right=True)`: boundary values belong to the lower category.
- Non-finite inputs (`NaN`, `±Infinity`) now return `NaN` so `stress_category` propagates alongside `utci` when `limit_inputs` rejects the input.
- Remove the now-unused `stress_categories_vals` array.

## Why this change is needed

`mapping()` was picking the numerically closest boundary value instead of classifying by range, which shifted every category transition to the midpoint of adjacent boundaries (26→32 transition moved to 29; inputs above 46 never returned "extreme heat stress"). Numerical `utci` was unaffected — only the `stress_category` string was wrong. Full repro in Issue #147.

## How I tested it

- [x] `npm test` — 1906 tests passed, 0 failed (37 new cases in `tests/models/utci.test.js` covering boundaries, midpoints, just-above transitions, and non-finite inputs)
- [x] `npm run build` — ESM + CJS succeeded
- [x] `npx prettier --write .` — no files rewritten

## Notes for reviewers

- `mapping()` is now exported from `utci.js` so tests can exercise it directly. Not added to `src/models/index.js`, so the top-level models registry is unchanged.
- `mapping_arr()` is untouched; it delegates to `mapping()` and inherits the fix for finite values. Cleanup is out of scope here.
- Public `utci()` signature is unchanged; only the returned `stress_category` string is corrected.

Closes #147

### DIFF
--- a/src/models/utci.js
+++ b/src/models/utci.js
@@ -23,10 +23,6 @@ const stress_categories = [
   "extreme heat stress",
 ];
 
-const stress_categories_vals = [
-  -40.0, -27.0, -13.0, 0.0, 9.0, 26, 32, 38, 46, 1000,
-];
-
 /**
  * Determines the Universal Thermal Climate Index (UTCI). The UTCI is the
     equivalent temperature for the environment derived from a reference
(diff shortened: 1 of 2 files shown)


TIAN-TOM:
@FedericoTartarini ready for review when you have time

Pyqren:
After rebasing and testing, I can confirm that the fix is working in the dashboard.

FedericoTartarini (approved):

PR #158, opened 2026-04-21
fix(two_nodes): initialise nSimulation to 1

TIAN-TOM:
Starts `nSimulation` in `two_nodes` at `1` to match pythermalcomfort's `n_simulation`, so both implementations run the same 59 inner iterations.

Fixes #153.

## What

In `src/models/two_nodes.js`, change the `nSimulation` initial value from `0` to `1`. The `while`/increment structure is unchanged; after this one-line edit, the loop matches `pythermalcomfort/models/two_nodes_gagge.py` (`n_simulation = 1`, `while n_simulation < 60`, `n_simulation += 1`).

## Why

Row 7 of `ts_two_nodes_gagge.json` (inputs `tdb=45, tr=45, v=1.1, rh=20, met=3, clo=0.2`) fails on five fields against the shared validation dataset because JS runs one more convergence iteration than the Python reference. Issue #153 has the before/after table and the Row 6 note.

## How tested

- Row 7 values from calling `two_nodes` directly with the fix, against `ts_two_nodes_gagge.json` tolerances: `e_rsw` 244.25 vs 244.9 (tol 2), `e_skin` 253.12 vs 253.1 (tol 4), `m_rsw` 359.19 vs 359.2 (tol 6), `pmv_set` 2.24 vs 2.2 (tol 0.1), `disc` 5.34 vs 5.3 (tol 0.1) — all within tolerance
- Full Jest suite (`npm test`) passes, including the existing `two_nodes.test.js` harness meta-tests; row-level coverage of the validation dataset depends on the harness repair tracked in #154
- `npm run build` clean

Row 6's separate `disc` divergence remains tracked in #152.

PR #162, opened 2026-04-27
fix(two_nodes): correct disc formula and refactor test harness

### LINKED ISSUE #135: Some Models Failing Tests in the Latest Build

### LINKED ISSUE #152: two_nodes: disc field diverges substantially on Row 6 (expected 6.0, actual ~16.56)

TIAN-TOM:
Row 6 of the two_nodes fixture was failing because the JS `disc` expression diverged from the Gagge reference. This PR realigns the JS expression with the reference and tidies up the harness to match the newer test files in the repo (`test.each` + the shared `validateResult` helper).

The pre-conditions for the fix are already on `main`: #145 corrected the `w_max` kwarg handling and #158 fixed `nSimulation`. With those in place the disc formula is the last remaining gap.

Tested locally with `npm test` on Node 22.x (68 suites, 2140 tests, all green). The harness commit alone reproduces the Row 6 failure and the fix commit clears it, which keeps the PR bisect-friendly.

Closes #152.
Closes #135 (the two_nodes part — pet_steady and phs were already fixed by #138).
Supersedes #154.

### DIFF
--- a/tests/models/two_nodes.test.js
+++ b/tests/models/two_nodes.test.js
@@ -1,73 +1,89 @@
-import { expect, describe, it, test, beforeAll } from "@jest/globals";
-import { loadTestData } from "./testUtils";
+import { expect, describe, test, it } from "@jest/globals";
+import { loadTestData, validateResult } from "./testUtils";
 import { two_nodes } from "../../src/models/two_nodes";
 import { testDataUrls } from "./comftest"; // Import all test URLs from comftest.js
 
 // Use the URL from comftest.js to fetch data for two_nodes tests
 const testDataUrl = testDataUrls.twoNodes; // Ensure the correct URL is added in comftest.js
 
-let testData;
-let tolerance;
+// Load data at module scope so test.each can register one test per row.
+const { testData, tolerances } = await loadTestData(testDataUrl, false);
 
-// Load data before tests
-beforeAll(async () => {
-  const result = await loadTestData(testDataUrl, "twoNodes"); // Load data from URL using loadTestData
-  testData = result.testData;
-  tolerance = result.tolerance || 0.1; // Set a default tolerance
-});
-
-// General test function to verify results for different test functions
-function runTest(testFunction, inputs, expected) {
-  const {
-    tdb,
-    tr,
-    v,
-    rh,
-    met,
-    clo,
-    wme,
-    body_surface_area,
-    p_atmospheric,
-    body_position,
-    max_skin_blood_flow,
-    kwargs,
-  } = inputs;
-
-  const result = testFunction(
-    tdb,
-    tr,
-    v,
-    rh,
-    met,
-    clo,
-    wme,
-    body_surface_area,
-    p_atmospheric,
-    body_position,
-    max_skin_blood_flow,
-    kwargs,
-  );
-
-  // Use specified tolerance or default to 0.0001
-  const tol = tolerance !== undefined ? tolerance : 0.0001;
-  expect(Math.abs(result - expected)).toBeLessThanOrEqual(tol);
-}
+// Skip rows with missing outputs or array-valued inputs; keep original
+// dataset index so failures are reported as "row N" matching the JSON file.
+const scalarRows = testData.data
+  .map((row, index) => ({ ...row, index }))
+  .filter(({ inputs, outputs }) => {
+    if (outputs === undefined || outputs === null) return false;
+    return !Object.values(inputs).some((value) => Array.isArray(value));
+  });
 
 describe("two_nodes related tests", () => {
-  it("should run two_nodes tests after data is loaded", () => {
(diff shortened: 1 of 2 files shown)


tests/models/two_nodes.test.js
```diff
@@ -1,73 +1,89 @@
...
+      clo,
+      wme,
+      body_surface_area,
+      p_atmospheric,
+      body_position,
+      max_skin_blood_flow,
+      mergedKwargs,
+    );
+
+    const resultSnakeCase = {
```
FedericoTartarini: which snake?

FedericoTartarini (changes requested):

tests/models/two_nodes.test.js
TIAN-TOM: Renamed this to `actualOutputs` so it describes the validation role rather than only the key format.

TIAN-TOM:
@FedericoTartarini Could you re-review when you have a chance?

FedericoTartarini (approved):

PR #163, opened 2026-04-27
feat(utilities): add maskOutOfRange helper for post-validation range masking

TIAN-TOM:
Adds a shared `maskOutOfRange(params, schema)` helper to `src/utilities/utilities.js` to support the range-check stage tracked in #142.

The helper walks a `{ paramName: { min, max } }` schema and returns a shallow-copied params object with each finite numeric value outside its `[min, max]` range replaced by `NaN` (per-element for arrays). The numeric range masking mirrors pythermalcomfort's `valid_range`. Inputs are not mutated.

**Positioning.** Intended as the post-validation range mask in the model entry path. Callers should pass each input through `validateInputs` first (the PR #145 type gate, which throws `TypeError` on non-numeric or non-finite inputs) and then call `maskOutOfRange` to collapse out-of-range numeric values to `NaN`. This matches the `TypeError` for invalid types / `NaN` for out-of-range split documented in `CLAUDE.md` §11.

**Direct utility callers.** Tests, debug scripts, and ad-hoc tooling that bypass `validateInputs` get a secondary defensive layer: each value passes through `Number.isFinite` before the range comparison, so non-numbers (`"25"`, `true`, `null`) and `Infinity`/`NaN` collapse to `NaN` without coercion. On the model entry path this branch is unreachable because `validateInputs` has already thrown; the guard exists for direct utility use only.

**Schema scope.** Only list a key in `schema` if you want it range-masked. An omitted optional caller param that arrives as `undefined` will collapse to `NaN` if the same key is also in `schema`, which is rarely the desired outcome — keep optional params out of the schema unless that collapse is intentional.

Params keys without schema rules are left unchanged; schema keys missing from params are skipped without adding new keys.

Models tracked in #142 need a uniform range guard before computation, and centralising the logic avoids per-model drift.

Scope is helper-only. No model is wired in this PR; model wiring will land in follow-up PRs to keep each change reviewable.

Tested locally with `npm test` on Node 22.x — 68 suites / 2145 tests pass. Unit tests cover scalar in/out-of-range, array per-element, `NaN` passthrough at scalar and array level, schema/params key gaps, input non-mutation, inclusive `[min, max]` boundaries, and the `Number.isFinite` guard for direct utility callers (strings, booleans, `null`, `undefined`, `Infinity`).

Refs #142.

### DIFF
--- a/src/utilities/index.js
+++ b/src/utilities/index.js
@@ -9,6 +9,7 @@ import {
   clo_typical_ensembles,
   clo_individual_garments,
   check_standard_compliance_array,
+  maskOutOfRange,
 } from "./utilities.js";
 
 /**
@@ -27,6 +28,7 @@ export {
   clo_typical_ensembles,
   clo_individual_garments,
   check_standard_compliance_array,
+  maskOutOfRange,
 };
 
 export default {
--- a/src/utilities/utilities.js
+++ b/src/utilities/utilities.js
@@ -928,3 +928,73 @@ export const validateInputs = (params, schema) => {
     }
   }
 };
+
+/**
+ * Replaces values that fall outside their declared `[min, max]` range with
+ * `NaN`, mirroring the range-check stage of pythermalcomfort's `valid_range`.
+ *
+ * Intended as the post-validation range mask in the model entry path. Callers
+ * should pass each input through `validateInputs` first (the PR #145 type
+ * gate, which throws `TypeError` for non-numeric or non-finite values) and
+ * then call this helper to collapse out-of-range numeric values to `NaN`.
+ *
+ * Scalars are replaced wholesale; arrays are checked per element. Scope is
+ * per-parameter only — cross-parameter rules, for example ASHRAE's
+ * `airspeed_control` cross-field check, are not handled here and remain the
+ * caller's responsibility (see `check_standard_compliance`).
+ *
+ * Direct utility callers (tests, debug, ad-hoc tooling) that bypass
+ * `validateInputs` get a secondary defensive layer: each value passes through
+ * `Number.isFinite` before the range comparison, so non-numbers (`"25"`,
+ * `true`, `null`) and `Infinity`/`NaN` collapse to `NaN` without coercion.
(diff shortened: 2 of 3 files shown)


TIAN-TOM:
@JitaoFu @Pyqren could you please take a peer-review pass when you have time? This PR only adds the validateRanges helper for #142; model wiring will be follow-up PRs.

Pyqren:
If I'm understanding this PR correctly, it's adding an additional validation tier (`maskOutOfRange`) to numeric inputs that replaces out-of-range numerics with `NaN`. No model calls `maskOutOfRange` yet, by design, since this will be in a future PR.

I applied the PR to a local branch and can confirm `maskOutOfRange` handles both scalars and arrays. No regressions found in the rest of the library.

JitaoFu:
### `maskOutOfRange` — Possible Overlap with Existing Implementation
**After reviewing how Python implements range checks, there are three approaches used:**

1. `check_standard_compliance` — shared cross-model validation based on standard
2. Direct `valid_range` calls— single-parameter range checks
3. Hard-coded — strict input validation

**The JS implementation already covers the first two:** 
- `check_standard_compliance` is fully ported and mirrors the Python version after PR
#165 
- all parameters checked via `valid_range` in Python are also checked in JS. A JS version of `valid_range` already exists in the codebase as well.

**The only remaining gap is the hard-coded validation (approach 3)**

This leaves `maskOutOfRange` with no clear use case. Could you clarify what this function is intended to cover? Also, the array support inside seems inconsistent with our team decision to drop array inputs.

JitaoFu:
I've added a range check parity review in the issue #142 . We could start addressing the gaps found there as a next step.

TIAN-TOM:
Closing this in favour of focusing the validation work on #142.

@JitaoFu's parity review there makes a clearer case for the actual gaps, and adding maskOutOfRange ahead of those PRs risks overlapping with the approaches he identified there.

If post-validation masking turns out to be needed during the gap PRs, happy to revisit then.

Thanks @Pyqren and @JitaoFu for the careful review.

PR #166, opened 2026-04-28
fix: remove premature rounding in Adaptive model (ASHRAE & EN)

Pyqren:
This PR addresses discrepancies in the ASHRAE and EN Adaptive model calculations, specifically regarding the "cliff-drop" behavior when applying elevated air speed.

**Changes:**
- Removed premature rounding of the neutral comfort temperature to one decimal place. 
- Updated the logic to only apply the cooling effect boost when the unadjusted base upper limit of the comfort zone is above 25°C. This aligns the library's behavior with the official Berkeley comfort tool, and prevents incorrect "Comfortable" results in regions where the adaptive cooling effect should not yet be applicable.
- Ensured rounding is still present in final return block of functions.

---

**Example Case (ASHRAE 55):**

With a Prevailing Mean Outdoor Temperature of 15.1°C, the unrounded base 90% limit is 24.981°C.

**Previous Behavior**: The library rounded  to 22.5°C, pushing the limit to exactly 25.0°C. For an operative temperature of 25.0°C, this incorrectly returned "Comfortable."

**New Behavior**: The unrounded base limit (24.981°C) is correctly identified as being ≤ 25 °C. Consequently, the cooling effect boost is not triggered, and an operative temperature of 25.0°C correctly returns "Too Warm".

---

This PR will ensure that the comfort tool's Adaptive model behavior mirrors the existing tool, including edge/corner cases.

### DIFF
--- a/src/models/adaptive_ashrae.js
+++ b/src/models/adaptive_ashrae.js
@@ -88,6 +88,7 @@ const ADAPTIVE_ASHRAE_SCHEMA = {
   v: { type: "number" },
   units: { enum: ["SI", "IP"] },
   limit_inputs: { type: "boolean" },
+  round_output: { type: "boolean", required: false },
 };
 
 export function adaptive_ashrae(
@@ -97,9 +98,23 @@ export function adaptive_ashrae(
   v,
   units = "SI",
   limit_inputs = true,
+  kwargs = {},
 ) {
+  const default_kwargs = {
+    round_output: true,
+  };
+  kwargs = Object.assign(default_kwargs, kwargs);
+
   validateInputs(
-    { tdb, tr, t_running_mean, v, units: units.toUpperCase(), limit_inputs },
+    {
+      tdb,
+      tr,
+      t_running_mean,
+      v,
+      units: units.toUpperCase(),
+      limit_inputs,
+      round_output: kwargs.round_output,
+    },
     ADAPTIVE_ASHRAE_SCHEMA,
   );
 
@@ -118,9 +133,7 @@ export function adaptive_ashrae(
     }));
   }
   const to = t_o(tdb, tr, v, standard);
-  // calculate cooling effect (ce) of elevated air speed when top > 25 degC.
-  const ce = get_ce(v, to);
-  // Relation between comfort and outdoor temperature
+
   let t_cmf = 0.31 * t_running_mean + 17.8;
 
   if (limit_inputs) {
@@ -144,12 +157,23 @@ export function adaptive_ashrae(
       t_cmf = NaN;
   }
 
-  t_cmf = round(t_cmf, 1);
+  if (kwargs.round_output) {
+    t_cmf = round(t_cmf, 1);
+  }
 
   let tmp_cmf_80_low = t_cmf - 3.5;
   let tmp_cmf_90_low = t_cmf - 2.5;
-  let tmp_cmf_80_up = t_cmf + 3.5 + ce;
-  let tmp_cmf_90_up = t_cmf + 2.5 + ce;
+  let tmp_cmf_80_up = t_cmf + 3.5;
+  let tmp_cmf_90_up = t_cmf + 2.5;
+
+  if (kwargs.round_output) {
+    const ce = get_ce(v, to);
+    tmp_cmf_80_up += ce;
+    tmp_cmf_90_up += ce;
+  } else {
+    tmp_cmf_80_up += get_ce(v, tmp_cmf_80_up);
+    tmp_cmf_90_up += get_ce(v, tmp_cmf_90_up);
+  }
 
   const acceptability_80 = tmp_cmf_80_low <= to && to <= tmp_cmf_80_up;
   const acceptability_90 = tmp_cmf_90_low <= to && to <= tmp_cmf_90_up;
@@ -171,6 +195,18 @@ export function adaptive_ashrae(
       },
       "SI",
     ));
+  }
+
+  if (kwargs.round_output) {
+    return {
+      tmp_cmf: round(t_cmf, 1),
+      tmp_cmf_80_low: round(tmp_cmf_80_low, 1),
+      tmp_cmf_80_up: round(tmp_cmf_80_up, 1),
+      tmp_cmf_90_low: round(tmp_cmf_90_low, 1),
+      tmp_cmf_90_up: round(tmp_cmf_90_up, 1),
+      acceptability_80,
+      acceptability_90,
+    };
   }
 
   return {
--- a/src/models/adaptive_en.js
+++ b/src/models/adaptive_en.js
@@ -75,6 +75,7 @@ const ADAPTIVE_EN_SCHEMA = {
   v: { type: "number" },
   units: { enum: ["SI", "IP"] },
   limit_inputs: { type: "boolean" },
+  round_output: { type: "boolean", required: false },
 };
 
 export function adaptive_en(
@@ -84,9 +85,23 @@ export function adaptive_en(
   v,
   units = "SI",
   limit_inputs = true,
+  kwargs = {},
 ) {
+  const default_kwargs = {
+    round_output: true,
+  };
+  kwargs = Object.assign(default_kwargs, kwargs);
+
   validateInputs(
-    { tdb, tr, t_running_mean, v, units: units.toUpperCase(), limit_inputs },
+    {
+      tdb,
+      tr,
+      t_running_mean,
+      v,
+      units: units.toUpperCase(),
+      limit_inputs,
+      round_output: kwargs.round_output,
+    },
     ADAPTIVE_EN_SCHEMA,
   );
 
@@ -103,8 +118,6 @@ export function adaptive_en(
 
   const to = t_o(tdb, tr, v, standard);
 
-  const ce = get_ce(v, to);
-
   let t_cmf = 0.33 * t_running_mean + 18.8;
 
   if (limit_inputs) {
@@ -115,9 +128,21 @@ export function adaptive_en(
   let t_cmf_i_lower = t_cmf - 3.0;
   let t_cmf_ii_lower = t_cmf - 4.0;
   let t_cmf_iii_lower = t_cmf - 5.0;
-  let t_cmf_i_upper = t_cmf + 2.0 + ce;
-  let t_cmf_ii_upper = t_cmf + 3.0 + ce;
-  let t_cmf_iii_upper = t_cmf + 4.0 + ce;
+
+  let t_cmf_i_upper = t_cmf + 2.0;
+  let t_cmf_ii_upper = t_cmf + 3.0;
+  let t_cmf_iii_upper = t_cmf + 4.0;
+
+  if (kwargs.round_output) {
+    const ce = get_ce(v, to);
+    t_cmf_i_upper += ce;
+    t_cmf_ii_upper += ce;
+    t_cmf_iii_upper += ce;
+  } else {
+    t_cmf_i_upper += get_ce(v, t_cmf_i_upper);
+    t_cmf_ii_upper += get_ce(v, t_cmf_ii_upper);
+    t_cmf_iii_upper += get_ce(v, t_cmf_iii_upper);
+  }
 
   const acceptability_i = t_cmf_i_lower <= to && to <= t_cmf_i_upper;
   const acceptability_ii = t_cmf_ii_lower <= to && to <= t_cmf_ii_upper;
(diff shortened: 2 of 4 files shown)


TIAN-TOM:
Hi @Pyqren, I tried the new logic on the example with `Prevailing Mean = 15.1` and want to flag something before merge.

`adaptive_ashrae(25, 25, 15.1, 0.6)` now returns `acceptability_90: false` and `tmp_cmf_90_up: 25.0`. The same inputs in pythermalcomfort 3.9.1 give `acceptability_90: true` and `tmp_cmf_90_up: 26.2`. `main` matches pythermalcomfort.

Looking at pythermalcomfort's implementation, `ce` is gated once on `(v >= 0.6) & (to >= 25)` and added unconditionally to all upper bounds. ASHRAE also rounds `t_cmf` to one decimal place before deriving bounds. This PR changes the gate so each `tmp_cmf_*_up > 25` is checked individually after derivation, and removes the pre-rounding for ASHRAE.

Mechanism for the divergence: with `t_running_mean = 15.1`, pythermalcomfort pre-rounds `t_cmf` to 22.5, giving `22.5 + 2.5 + 1.2 = 26.2` (boost applied, comfortable). This PR keeps `t_cmf = 22.481`, giving `22.481 + 2.5 = 24.981`, which fails the new `> 25` gate, so the boost is not applied and the final value rounds to 25.0.

If this divergence is intentional, it would be worth adding tests that lock in the new contract. Otherwise, keeping the pythermalcomfort `ce` gate would keep this implementation aligned with the reference behaviour.

Verified by running PR head and pythermalcomfort 3.9.1 locally on the same input.

Pyqren:
@TIAN-TOM 

Thanks for reviewing this, Tom.

The divergence is indeed intentional to achieve parity with the official [CBE Thermal Comfort Tool](https://comfort.cbe.berkeley.edu/)

My testing confirmed that the CBE tool handles these edge cases differently (and more precisely) than `pythermalcomfort`:

- The CBE tool (and my implementation) triggers the air speed boost based on whether the base comfort boundary is above 25.0°C, rather than just checking if the operative temperature is above 25.0°C.
- The CBE tool uses high-precision calculations for these thresholds. For example, at outdoor temperature = 15.1°C, it correctly identifies that the base limit is slightly below 25.0°C (approx 24.98°C) and therefore does not apply the boost. `pythermalcomfort`'s pre-rounding of comfort temperature loses this detail, leading to the incorrect "Comfortable" result that we had before.

To address this, I have:

- Added detailed comments to `adaptive_ashrae.js` and `adaptive_en.js` explaining the use of high-precision logic and per-boundary gating for CBE tool parity. This prevents future accidental reverts to the `pythermalcomfort` behavior.
- Created a new test suite, `adaptive_precision.test.js`, which explicitly tests these edge cases (e.g. ASHRAE 15.1/25.1/0.6) and asserts the CBE-aligned behavior as well as lock in the new logic.

I've introduced another [commit](https://github.com/FedericoTartarini/jsthermalcomfort/pull/166/commits/0b8d466c8588ca9e8e63ec17a2465e867d821a81). All 38 tests are passing on my end. Please check and confirm.

TIAN-TOM:
Hi @Pyqren, thanks for the docs and tests. The PR's intent reads clearly now: it locks parity with the CBE Tool, not pythermalcomfort.

@FedericoTartarini might want to weigh in here, since this is more about which reference to validate against than about the code itself. The repo's been validating adaptive against pythermalcomfort (via the `validation-data-comfort-models` submodule), so worth a quick call on whether CBE Tool parity is the way to go for adaptive.

Tested locally on `0b8d466`. ASHRAE output is the same as before (`tmp_cmf_90_up: 25.0`, `acceptability_90: false`). EN matches pythermalcomfort on this input.

src/models/adaptive_ashrae.js
```diff
@@ -144,12 +144,26 @@ export function adaptive_ashrae(
...
+  // to ensure high-precision threshold triggers for air speed cooling effects.
+  // This matches the official CBE Thermal Comfort Tool's behavior.
+  // Previous versions and pythermalcomfort 3.9.1 pre-round t_cmf to 1 decimal place.
+  // t_cmf = round(t_cmf, 1);
 
   let tmp_cmf_80_low = t_cmf - 3.5;
   let tmp_cmf_90_low = t_cmf - 2.5;
-  let tmp_cmf_80_up = t_cmf + 3.5 + ce;
-  let tmp_cmf_90_up = t_cmf + 2.5 + ce;
+  let tmp_cmf_80_up = t_cmf + 3.5;
```
FedericoTartarini: Are you sure this is correct? I would not fix it here but in the calculation of the cooling effect. The cooling effect should be calculated as a function of the wind speed and here we should just be using the old code in which we add the cooling effect if needed.

src/models/adaptive_ashrae.js
```diff
@@ -144,12 +144,26 @@ export function adaptive_ashrae(
       t_cmf = NaN;
   }
 
-  t_cmf = round(t_cmf, 1);
+  // The comfort temperature (t_cmf) is kept unrounded during boundary calculations
```
FedericoTartarini: This comment should not be added to the code, but to the pull request description

src/models/adaptive_ashrae.js
```diff
@@ -174,11 +188,11 @@ export function adaptive_ashrae(
   }
 
   return {
-    tmp_cmf: t_cmf,
-    tmp_cmf_80_low,
-    tmp_cmf_80_up,
-    tmp_cmf_90_low,
-    tmp_cmf_90_up,
+    tmp_cmf: round(t_cmf, 1),
```
FedericoTartarini: This is not correct, we should round the output only if the user wants it to be rounded the same as we have done in the PMV model.

src/models/adaptive_en.js
```diff
@@ -115,9 +115,21 @@ export function adaptive_en(
   let t_cmf_i_lower = t_cmf - 3.0;
   let t_cmf_ii_lower = t_cmf - 4.0;
   let t_cmf_iii_lower = t_cmf - 5.0;
-  let t_cmf_i_upper = t_cmf + 2.0 + ce;
-  let t_cmf_ii_upper = t_cmf + 3.0 + ce;
-  let t_cmf_iii_upper = t_cmf + 4.0 + ce;
+  // Boundaries are gated individually against the 25.0°C threshold
+  // using unrounded t_cmf to ensure precise cliff-drop behavior
+  // consistent with the CBE Thermal Comfort Tool.
+  let t_cmf_i_upper = t_cmf + 2.0;
```
FedericoTartarini: See my previous comment.

FedericoTartarini (changes requested):

src/models/adaptive_en.js
Pyqren: @FedericoTartarini Hi Federico, just for clarity, are you referring to [this comment](https://github.com/FedericoTartarini/jsthermalcomfort/pull/166#discussion_r3158577443) about calculation of the cooling effect?

FedericoTartarini:
> Hi @Pyqren, thanks for the docs and tests. The PR's intent reads clearly now: it locks parity with the CBE Tool, not pythermalcomfort.
> 
> @FedericoTartarini might want to weigh in here, since this is more about which reference to validate against than about the code itself. The repo's been validating adaptive against pythermalcomfort (via the `validation-data-comfort-models` submodule), so worth a quick call on whether CBE Tool parity is the way to go for adaptive.
> 
> Tested locally on `0b8d466`. ASHRAE output is the same as before (`tmp_cmf_90_up: 25.0`, `acceptability_90: false`). EN matches pythermalcomfort on this input.

@TIAN-TOM great point. @Pyqren jsthermalcomfort should have parity with pythermalcomfort, not the CBE thermal comfort tool. However, said that, I think in PyThermalComfort we should be implementing the same functionality to adjust the upper threshold based on the wind speed since this is part of the Standard now. We should fix this pull request and then also implement the changes in PyThermalComfort.

src/models/adaptive_en.js
FedericoTartarini: I'm basically saying that there is a function that calculates the cooling effect and this function should return 0 if we are below a certain temperature threshold. We should not be splitting the checks on the cooling effect into different parts.

When I wrote my previous comment I meant that we should not have comments in line in the code unless strictly necessary and we should have a consistent way in which we add the cooling effect to the upper comfort threshold

FedericoTartarini:
@Pyqren please also refer to this discussion https://github.com/CenterForTheBuiltEnvironment/comfort_tool/issues/89#issuecomment-4336956956

src/models/adaptive_ashrae.js
Pyqren: I'll consolidate the cooling effect threshold logic into `cooling_effect.js` and the `get_ce` helper instead.

src/models/adaptive_ashrae.js
Pyqren: Will remove comments.

src/models/adaptive_ashrae.js
Pyqren: Added a `round_output` parameter to allow opting into the high-precision logic.

src/models/adaptive_en.js
Pyqren: Noted. I've restored the parity with `pythermalcomfort`, including early rounding and gating, and consolidate the cooling effect threshold logic into `cooling_effect.js`.

FedericoTartarini:
@Pyqren may I know why you are opening and closing it? Sorry, I'm just asking because I keep receiving emails and I would like to understand what I need to do.

Pyqren:
> @Pyqren may I know why you are opening and closing it? Sorry, I'm just asking because I keep receiving emails and I would like to understand what I need to do.

@FedericoTartarini My apologies, I closed the PR by mistake and have reopened it after submitting a new commit.

I've restored the `pythermalcomfort` parity and made adjustments as per your comments. Please see my replies to the comments for more details. The comfort-tool's own `adaptive.js` will ensure the chart is accurate.

src/models/adaptive_ashrae.js
```diff
@@ -97,9 +98,23 @@ export function adaptive_ashrae(
   v,
   units = "SI",
   limit_inputs = true,
+  kwargs = {},
```
FedericoTartarini, line 101: Can't round_output be a simple input rather than using the kwargs argument and making the code more complicated?

src/models/adaptive_ashrae.js
FedericoTartarini: The cooling effect function is something completely different and should not be mixed with the adaptive cooling effect. The cooling effect is there to calculate the cooling provided by a fan, but the helper get CE is supposed to be much more lean and simple. For the adaptive model we just need to apply a constant increase to the temperature.

src/models/adaptive_ashrae.js
```diff
@@ -144,12 +157,23 @@ export function adaptive_ashrae(
...
+  }
 
   let tmp_cmf_80_low = t_cmf - 3.5;
   let tmp_cmf_90_low = t_cmf - 2.5;
-  let tmp_cmf_80_up = t_cmf + 3.5 + ce;
-  let tmp_cmf_90_up = t_cmf + 2.5 + ce;
+  let tmp_cmf_80_up = t_cmf + 3.5;
+  let tmp_cmf_90_up = t_cmf + 2.5;
+
+  if (kwargs.round_output) {
```
FedericoTartarini, line 169: I don't think we should be repeating the code here and getting the cooling effect twice. We should just simply get the cooling effect and then have another if statement for the rounding. Each part of the code should just do one thing.

src/models/adaptive_ashrae.js
```diff
@@ -144,12 +157,23 @@ export function adaptive_ashrae(
...
+    const ce = get_ce(v, to);
```
FedericoTartarini, line 170: I'm actually very confused about what this code is doing. Perhaps you can ignore my previous comment but I don't understand what we are trying to do here with an else statement.

src/models/adaptive_en.js
```diff
@@ -84,9 +85,23 @@ export function adaptive_en(
   v,
   units = "SI",
   limit_inputs = true,
+  kwargs = {},
```
FedericoTartarini, line 88: See my previous comment about removing this.

src/models/adaptive_en.js
```diff
@@ -115,9 +128,21 @@ export function adaptive_en(
...
   let t_cmf_iii_lower = t_cmf - 5.0;
-  let t_cmf_i_upper = t_cmf + 2.0 + ce;
-  let t_cmf_ii_upper = t_cmf + 3.0 + ce;
-  let t_cmf_iii_upper = t_cmf + 4.0 + ce;
+
+  let t_cmf_i_upper = t_cmf + 2.0;
+  let t_cmf_ii_upper = t_cmf + 3.0;
+  let t_cmf_iii_upper = t_cmf + 4.0;
+
+  if (kwargs.round_output) {
```
FedericoTartarini, line 136: I don't understand what is the point of this if statement. We should not be rounding the values here.

src/models/cooling_effect.js
```diff
@@ -76,12 +76,20 @@ export function cooling_effect(
...
   units = "SI",
+  threshold = undefined,
+  temp_checked = undefined,
 ) {
   validateInputs(
     { tdb, tr, vr, rh, met, clo, wme, units: units.toUpperCase() },
     COOLING_EFFECT_SCHEMA,
   );
 
+  if (threshold !== undefined && temp_checked !== undefined) {
```
FedericoTartarini, line 87: Could you please check "PyThermalComfort" and see how we are handling it, the cooling effect over there As you can see here, this function is not related at all with the adaptive model because it is accepting a clotting metabolic rate as an input.

FedericoTartarini (changes requested):
I think that this pull request has a few issues that needs to be solved. We need to have code parity with the PyTermalComfort and we should make sure that we are not mixing the two cooling effect functions.

Pyqren:
Looking further into this and `comfort-tool`, it seems that a JS-only fix in `jsthermalcomfort` cannot both handle the edge-case result and preserve parity with current PyThermalComfort. I believe there are two approaches:

- **Approach 1**: Implement the fix entirely in `comfort-tool`, with [CBE Tool's PR#91 ](https://github.com/CenterForTheBuiltEnvironment/comfort_tool/pull/91) as blueprint. No changes to `jsthermalcomfort`.
- **Approach 2**: For a `jsthermalcomfort `fix while preserving parity, PyThermalComfort would first adopt this per-boundary >=25 C behavior, then jsthermalcomfort would mirror it exactly.

In either cases, I think it would be worth it to close this PR.

@FedericoTartarini: Please let me know which approach is preferred, or if you'd like to suggest another approach.

FedericoTartarini:
> Looking further into this and `comfort-tool`, it seems that a JS-only fix in `jsthermalcomfort` cannot both handle the edge-case result and preserve parity with current PyThermalComfort. I believe there are two approaches:
> 
> * **Approach 1**: Implement the fix entirely in `comfort-tool`, with [CBE Tool's PR#91 ](https://github.com/CenterForTheBuiltEnvironment/comfort_tool/pull/91) as blueprint. No changes to `jsthermalcomfort`.
> * **Approach 2**: For a `jsthermalcomfort `fix while preserving parity, PyThermalComfort would first adopt this per-boundary >=25 C behavior, then jsthermalcomfort would mirror it exactly.
> 
> In either cases, I think it would be worth it to close this PR.
> 
> @FedericoTartarini: Please let me know which approach is preferred, or if you'd like to suggest another approach.

I would like to make sure that we fix the issue in both packages. I think that if we have an issue we should follow approach 2 and fix it in both packages. Let's discuss this today.

PR #168, opened 2026-05-03
Add relative humidity range check to humidex

TIAN-TOM:
## What this PR does

- Adds a `[0, 100]` range check on `rh` in `humidex()`.
- Out-of-range relative humidity now throws `RangeError` because it is physically invalid input, matching pythermalcomfort 3.9.3's `ValueError` behaviour.
- Adds 4 out-of-range throw cases (rh = -25, -0.1, 100.1, 150) and 2 boundary sanity cases (rh = 0, rh = 100).

## Why this change is needed

- Part of #142 (input validation across all models).
- pythermalcomfort 3.9.3 raises `ValueError` for `rh` outside `[0, 100]`. `RangeError` is the JavaScript analogue for value-domain violations on a valid type.
- Keeps `humidex` in the physically invalid input category rather than the applicability-limit category that returns `NaN` under `limit_inputs`.

## How I tested it

- [x] `npm test` (2150 -> 2156 tests in the original push; final branch adds 4 throw cases + 2 boundary cases)
- [x] `npm run build`
- [x] `node --experimental-vm-modules ./node_modules/.bin/jest --testPathPattern="humidex"` in a temporary worktree for the final head commit: 28 tests passed

## Notes for reviewers

- No fixture changes needed: every existing `humidex` validation case has rh in `[0, 100]`.
- The range check is inline and does not touch `validateInputs()`, so it should not conflict with #169's shared min/max schema work.
- Test file picks up `expect, test` in the import line; the new describe block was relying on implicit globals, so this is a tiny side cleanup, not a behaviour change.

### DIFF
--- a/src/models/humidex.js
+++ b/src/models/humidex.js
@@ -18,8 +18,8 @@ import { round, validateInputs } from "../utilities/utilities.js";
  * Humidex differs from the heat index in being related to the dew point
  * rather than relative humidity {@link #ref_15|[15]}.
  *
- * Relative humidity outside [0, 100] returns
- * `{ humidex: NaN, discomfort: NaN }`.
+ * Relative humidity outside [0, 100] is physically invalid and throws a
+ * `RangeError`.
  *
  * @public
  * @memberof models
@@ -30,6 +30,8 @@ import { round, validateInputs } from "../utilities/utilities.js";
  * @param {object} [options] - configuration options for the function.
  * @param {boolean} [options.round = true] - If true, rounds output value. If
  * false, it does not.
+ *
+ * @throws {RangeError} when `rh` is outside `[0, 100]`.
  *
  * @returns {HumidexResult} the result given the provided temperature and
  * relative humidity.
@@ -48,7 +50,7 @@ export function humidex(tdb, rh, options = { round: true }) {
   validateInputs({ tdb, rh, round: options.round }, HUMIDEX_SCHEMA);
 
   if (rh < 0 || rh > 100) {
-    return { humidex: NaN, discomfort: NaN };
+    throw new RangeError("Relative humidity must be between 0 and 100%");
   }
 
   let hi =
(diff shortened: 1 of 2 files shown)


JitaoFu:
Hi, @TIAN-TOM While working on the `phs` fix I realised we have been using 'range check' loosely as a catch-all term, which may have caused some misalignment. It's probably worth distinguishing between hard validation (physically invalid inputs that should throw) and compliance check (inputs outside a standard's applicability range that should return `NaN`) going forward.

**In Pythermal's range check there are two distinct cases:** 
- inputs that are physically invalid (e.g. negative weight, rh outside [0, 100]) raise a `ValueError`
- inputs that are valid but outside a standard's applicability range return `NaN`. 

I think we should mirror this in JS — physically invalid inputs should throw a `RangeError`, and out-of-range but physically valid inputs should return `NaN`. The `humidex` rh case falls into the first category, so it should throw rather than return NaN.

**On the implementation side**

`validateInputs()` in PR #169 already supports throwing RangeError for invalid inputs. However, if we both modify validateInputs at the same time we risk conflicts. What do you suggest？

TIAN-TOM:
@JitaoFu You're right about the two-category split. I checked pythermalcomfort 3.9.3 and `humidex(tdb=25, rh=-5)` raises `ValueError`, so `rh` outside `[0, 100]` is a physically invalid input. The JS analogue is `RangeError`.

I'll amend this PR shortly to throw `RangeError` for `rh` outside `[0, 100]` inline, without touching `validateInputs()`, so it should not conflict with #169.

Once #169 lands, we can refactor `humidex` to use the shared `min` / `max` schema rules if that still makes sense.

For the next `heat_index` work, I agree that the `tdb < 27` Rothfusz gate is an applicability limit rather than physical invalidity, so that should stay as `NaN` under `limit_inputs=true`.

Thanks for flagging the distinction.

JitaoFu:
Thanks for checking and for the clean fix — appreciated the thought on avoiding the conflict too. Agreed on the heat_index call.

FedericoTartarini (approved):

PR #165, opened 2026-04-27
refactor: remove remaining source-level *_array APIs

TIAN-TOM:
**What this PR does**

- Remove `check_standard_compliance_array` from `src/utilities/utilities.js` and from the utilities export.
- Migrate the four callers (`adaptive_ashrae`, `pmv_ppd`, `set_tmp`, `use_fans_heatwaves`) to the scalar `check_standard_compliance` helper.
- Port the ASHRAE `airspeed_control` cross-field check and a `FAN_HEATWAVES` branch onto the scalar helper so applicability gating is preserved.
- Fix a pre-existing operator-precedence bug in the ISO branch (`key === "v" || (key === "vr" && ...)`) that was masking the `v`/`vr` range check.
- Remove the ASHRAE `[-100, 100]` PMV output gate (port-era artifact not in pythermalcomfort 3.9.3) while preserving the ISO `[-2, 2]` PMV applicability bound from ISO 7730.
- Replace the array-form tests with scalar coverage for the ported branches.
- Remove `t_o_array` from `src/psychrometrics/t_o.js` (its only in-tree caller was the removed helper).
- Remove 18 stale `*_array` display-name keys from `docs_theme/index.js`.
- Remove three broken example files in `docs/examples/` (`calc_adaptive_EN.js`, `calc_pmv_ppd.js`, `calc_utci.js`) that imported array APIs removed in PR #141.

**Why this change is needed**

`check_standard_compliance_array` was the last `*_array` utility export remaining after PR #141. This PR removes it along with the residual references (`t_o_array`, docs theme keys, broken examples), so no `*_array` public export remains at the source level.

**How I tested it**

- [x] `npm test` — 68 suites, 2137 tests, all green
- [x] `npm run build`

**Notes for reviewers**

The ISO precedence fix and `FAN_HEATWAVES` branch are in this PR because the migration would otherwise change behaviour for `use_fans_heatwaves` and silently mask a `v`/`vr` validation gap.

Generated docs are not regenerated in this PR. A fresh `npm run docs` on this branch would also migrate the theme to a multi-page setup, which is unrelated to this cleanup. Docs regeneration is left for a separate pass.

Closes #159. Partially addresses #142.

PR #169, opened 2026-05-03
fix(phs): align range checks and compliance behaviour with Python

JitaoFu:
## What this PR does
**This PR addresses the phs gaps identified in the range check parity review:** https://github.com/FedericoTartarini/jsthermalcomfort/issues/142#issuecomment-4354355823

1. Extends `validateInputs` in `utilities.js` with three new range rules: `min`, `max`, and `exclusiveMin`, each throwing a `RangeError` on violation
2. Fixes two bugs in `_iso7933_compliance`: 
    - corrects the `tr` check from `tr - tdb ∈ [0, 60]` to `tr ∈ [0, 60]`
    - fixes a unit mismatch in the `p_a` check (kPa vs %), and updates the corresponding warning messages

3. Adds hard validation rules to the PHS schema for `acclimatized`, `weight`, `t_sk_t_cr_wg`, `sweat_rate_watt`, and `evap_load_wm2_min` to align input constraints with the pythermalcomfort library.
4. Adds `limit_inputs` support to `phs`: when `true` (default), out-of-range inputs trigger a warning and return all NaN; when `false`, compliance checks are skipped and calculation proceeds
5. Adds corresponding tests for all new validation rules and compliance fixes

## Why this change is needed
- The hard validation cases in `phs` (e.g. negative weight, `t_sk_t_cr_wg` outside `[0, 1]`) represent physically invalid inputs. Python raises `ValueError` for these; JS should throw `RangeError` rather than silently proceeding or returning NaN
- The `_iso7933_compliance` bugs caused incorrect compliance warnings: the `tr` check was unintentionally relative to `tdb`, and `p_a` was being compared in the wrong unit, meaning some out-of-range inputs passed through without warning
- `limit_inputs` support aligns `phs` behaviour with the Python reference implementation

## How I tested it
- [ ] npm test

## Notes for reviewers
1. The new `min`, `max`, and `exclusiveMin` rules in `validateInputs` are also needed by `humidex` (for the `rh` hard validation) and potentially other models. If those PRs are in flight, they can reuse this utility directly rather than implementing the range check inline.

2. **Out of scope: ISO 7933:2023 `p_a` lower bound (0 → 0.5 kPa)** 
The 2023 standard raises the `p_a` applicability lower limit from 0 to 0.5 kPa. This was identified during the Python parity audit but is intentionally excluded from this PR. As patching a single value in isolation risks producing an incomplete or incorrect 2023 compliance check. The correct approach is to audit all compliance parameters against the 2023 standard text and add full 2023 support in a dedicated PR.


TIAN-TOM:
@JitaoFu heads-up: #165 (the `*_array` cleanup) is queued ahead of this PR and will remove `check_standard_compliance_array`. When it merges, `tests/utilities/utilities.test.js` here will fail to load because of the existing import. Fix is one line: drop that name from the import list.

Nothing to do yet. I'll ping again once #165 actually lands.

FedericoTartarini (approved):

JitaoFu:
Hi @FedericoTartarini, the conflict has been resolved and all checks are passing. Ready to merge.

FedericoTartarini (approved):

PR #173, opened 2026-05-09
fix(phs): make compliance check model-aware for ISO 7933:2004 and 2023

### LINKED ISSUE #172: bug(phs): compliance check always uses ISO 7933:2004 limits regardless of model parameter
## Description

The compliance check for `phs` model lacks model-aware logic. Regardless of which model is passed to `phs()`, `_iso7933_compliance` always enforces ISO 7933:2004 rules — it recomputes `p_a` using the (2004 `p_sat()`-based) formula and applies the 2004 lower limit of `0 kPa` even when `model="7933-2023"` is selected.

## Expected behaviour

When `model="7933-2004"`:

1. Compute `p_a` using the (`p_sat()`-based) formula: `p_a = p_sat(tdb) / 1000 * rh / 100`
2. Validate `p_a` against the ISO 7933:2004 applicability limit: `0 ≤ p_a ≤ 4.5 kPa`
3. All warning messages should reference ISO 7933:2004

When `model="7933-2023"`:

1. Compute `p_a` using the Magnus formula: `p_a = 0.6105 * exp(17.27 * tdb / (tdb + 237.3)) * rh / 100`
2. Validate `p_a` against the ISO 7933:2023 applicability limit: `0.5 ≤ p_a ≤ 4.5 kPa`
3. All warning messages should reference ISO 7933:2023

In both cases, all other applicability limits (`tdb`, `tr`, `v`, `met`, `clo`) remain the same and are unaffected by the model selection.

## Reference

Python `pythermalcomfort` handles this correctly — it passes the pre-computed `p_a` (calculated with the model-appropriate formula) directly to the compliance check, and applies separate limits per standard (see `pythermalcomfort/utilities.py`, `_check_standard_compliance_array`).

### Option A — Shared function, parameter-distinguished

**phs.js**
```js
check_standard_compliance(
  model, {
  tdb, tr, v, p_a,
  met: met_watt * body_surface_area(...),
  clo
});
```

**utilities.js**
```js
case "7933-2004":
  return _iso7933_compliance(kwargs, "2004", 0);
case "7933-2023":
  return _iso7933_compliance(kwargs, "2023", 0.5);

function _iso7933_compliance(kwargs, year, p_a_min) {
  let warnings = [];
  if (kwargs.tdb > 50 || kwargs.tdb < 15)
    warnings.push(`ISO 7933:${year} air temperature applicability limits between 15 and 50 ºC`);
  if (kwargs.p_a > 4.5 || kwargs.p_a < p_a_min)
    warnings.push(`ISO 7933:${year} p_a applicability limits between ${p_a_min} and 4.5 kPa`);
  if (kwargs.tr > 60 || kwargs.tr < 0)
    warnings.push(`ISO 7933:${year} tr applicability limits between 0 and 60 ºC`);
  if (kwargs.v > 3 || kwargs.v < 0)
    warnings.push(`ISO 7933:${year} air speed applicability limits between 0 and 3 m/s`);
  if (kwargs.met > 450 || kwargs.met < 100)
    warnings.push(`ISO 7933:${year} met applicability limits between 100 and 450 W/m²`);
  if (kwargs.clo > 1 || kwargs.clo < 0.1)
    warnings.push(`ISO 7933:${year} clo applicability limits between 0.1 and 1 clo`);
  return warnings;
}
```

### Option B — Two independent functions, semantically explicit (mirrors Python)

**phs.js**
```js
check_standard_compliance(
  model, {
  tdb, tr, v, p_a,
  met: met_watt * body_surface_area(...),
  clo
});
```

**utilities.js**
```js
case "7933-2004":
  return _iso7933_2004_compliance(kwargs);
case "7933-2023":
  return _iso7933_2023_compliance(kwargs);

// based on ISO 7933:2004 Annex A
function _iso7933_2004_compliance(kwargs) {
  let warnings = [];
  if (kwargs.tdb > 50 || kwargs.tdb < 15)
    warnings.push("ISO 7933:2004 air temperature applicability limits between 15 and 50 ºC");
  if (kwargs.p_a > 4.5 || kwargs.p_a < 0)       // p_a lower limit: 0
    warnings.push("ISO 7933:2004 p_a applicability limits between 0 and 4.5 kPa");
  if (kwargs.tr > 60 || kwargs.tr < 0)
    warnings.push("ISO 7933:2004 tr applicability limits between 0 and 60 ºC");
  if (kwargs.v > 3 || kwargs.v < 0)
    warnings.push("ISO 7933:2004 air speed applicability limits between 0 and 3 m/s");
  if (kwargs.met > 450 || kwargs.met < 100)
    warnings.push("ISO 7933:2004 met applicability limits between 100 and 450 W/m²");
  if (kwargs.clo > 1 || kwargs.clo < 0.1)
    warnings.push("ISO 7933:2004 clo applicability limits between 0.1 and 1 clo");
  return warnings;
}

// based on ISO 7933:2023 Annex A
function _iso7933_2023_compliance(kwargs) {
  let warnings = [];
  if (kwargs.tdb > 50 || kwargs.tdb < 15)
    warnings.push("ISO 7933:2023 air temperature applicability limits between 15 and 50 ºC");
  if (kwargs.p_a > 4.5 || kwargs.p_a < 0.5)     // p_a lower limit: 0.5 (vs 0 in 2004)
    warnings.push("ISO 7933:2023 p_a applicability limits between 0.5 and 4.5 kPa");
  if (kwargs.tr > 60 || kwargs.tr < 0)
    warnings.push("ISO 7933:2023 tr applicability limits between 0 and 60 ºC");
  if (kwargs.v > 3 || kwargs.v < 0)
    warnings.push("ISO 7933:2023 air speed applicability limits between 0 and 3 m/s");
  if (kwargs.met > 450 || kwargs.met < 100)
    warnings.push("ISO 7933:2023 met applicability limits between 100 and 450 W/m²");
  if (kwargs.clo > 1 || kwargs.clo < 0.1)
    warnings.push("ISO 7933:2023 clo applicability limits between 0.1 and 1 clo");
  return warnings;
}
```
The only substantive difference between the two is the p_a lower limit: 0 for the 2004 standard vs 0.5 for the 2023 standard. Option A passes this difference as a parameter, keeping the code more concise; Option B uses two independent functions, making each version self-contained and easier to read without tracing parameter origins.

> Note: _iso7933_compliance is only called from phs.js — no other models are affected by this change.

JitaoFu:
#### **Logic fix**
- Split `_iso7933_compliance` into two model-aware functions: `_iso7933_2004_compliance` and `_iso7933_2023_compliance`, mirroring the structure in pythermalcomfort
- Move `p_a` calculation before the compliance check in `phs.js` so the pre-computed value is passed directly instead of being recomputed inside the compliance function
- Apply the correct standard limits for each model (ISO 7933:2023 requires `p_a >= 0.5 kPa`; ISO 7933:2004 uses `p_a >= 0`)

#### **Tests**
- Add tests to `phs.test.js` asserting that the same low-`p_a` input is rejected under 2023 but accepted under 2004
- Update `utilities.test.js` to test `check_standard_compliance` against `"7933-2004"` / `"7933-2023"` directly (replacing the old `"ISO7933"` key), and pass `p_a` as a pre-computed value instead of `rh`

## Why this change is needed

`_iso7933_compliance` previously always enforced ISO 7933:2004 rules regardless of the `model` parameter — it recomputed `p_a` using the 2004 `p_sat()`-based formula and applied a lower `p_a` limit of `0 kPa` even when `model="7933-2023"`, producing incorrect compliance results for the 2023 standard. 

See issue #172 

## How I tested it

- [x] `npm test` — 69 suites, 2258 tests passed, no regression found.

## Notes for reviewers

- The `p_a` pre-computation in `phs.js` is a prerequisite for correctly dispatching to the two compliance functions — the two changes are coupled.

Closes #172

### DIFF
--- a/src/utilities/utilities.js
+++ b/src/utilities/utilities.js
@@ -71,7 +71,19 @@ export function transpose_sharp_altitude(sharp, altitude) {
  */
 
 /**
- * @typedef {"ANKLE_DRAFT" | "ASHRAE" | "FAN_HEATWAVES" | "ISO" | "7933-2004" | "7933-2023"} Standard
+ * Standard names accepted by {@link check_standard_compliance}.
+ */
+export const Standard = Object.freeze({
+  ANKLE_DRAFT: "ANKLE_DRAFT",
+  ASHRAE: "ASHRAE",
+  FAN_HEATWAVES: "FAN_HEATWAVES",
+  ISO: "ISO",
+  ISO_7933_2004: "7933-2004",
+  ISO_7933_2023: "7933-2023",
+});
+
+/**
+ * @typedef {(typeof Standard)[keyof typeof Standard]} Standard
  */
 
 /**
@@ -84,17 +96,17 @@ export function transpose_sharp_altitude(sharp, altitude) {
  */
 export function check_standard_compliance(standard, kwargs) {
   switch (standard) {
-    case "ANKLE_DRAFT":
+    case Standard.ANKLE_DRAFT:
       return _ankle_draft_compliance(kwargs);
-    case "ASHRAE":
+    case Standard.ASHRAE:
       return _ashrae_compliance(kwargs);
-    case "FAN_HEATWAVES":
+    case Standard.FAN_HEATWAVES:
       return _fan_heatwaves_compliance(kwargs);
-    case "ISO":
+    case Standard.ISO:
       return _iso_compliance(kwargs);
-    case "7933-2004":
+    case Standard.ISO_7933_2004:
       return _iso7933_2004_compliance(kwargs);
-    case "7933-2023":
+    case Standard.ISO_7933_2023:
       return _iso7933_2023_compliance(kwargs);
     default:
       throw new Error("Unknown standard");
--- a/src/utilities/utilities.js
+++ b/src/utilities/utilities.js
@@ -71,7 +71,7 @@ export function transpose_sharp_altitude(sharp, altitude) {
  */
 
 /**
- * @typedef {"ANKLE_DRAFT" | "ASHRAE" | "FAN_HEATWAVES" | "ISO" | "ISO7933"} Standard
+ * @typedef {"ANKLE_DRAFT" | "ASHRAE" | "FAN_HEATWAVES" | "ISO" | "7933-2004" | "7933-2023"} Standard
  */
 
 /**
@@ -92,8 +92,10 @@ export function check_standard_compliance(standard, kwargs) {
       return _fan_heatwaves_compliance(kwargs);
     case "ISO":
       return _iso_compliance(kwargs);
-    case "ISO7933":
-      return _iso7933_compliance(kwargs);
+    case "7933-2004":
+      return _iso7933_2004_compliance(kwargs);
+    case "7933-2023":
+      return _iso7933_2023_compliance(kwargs);
     default:
       throw new Error("Unknown standard");
   }
@@ -275,17 +277,18 @@ function _iso_compliance(kwargs) {
  *
  * @returns {string[]} strings with warnings emitted
  */
-function _iso7933_compliance(kwargs) {
+function _iso7933_2004_compliance(kwargs) {
+  // based on ISO 7933:2004 Annex A
   if (
     kwargs.tdb === undefined ||
-    kwargs.rh === undefined ||
+    kwargs.p_a === undefined ||
     kwargs.tr === undefined ||
     kwargs.v === undefined ||
     kwargs.met === undefined ||
     kwargs.clo === undefined
   ) {
     throw new Error(
-      `Missing arguments for ISO7933 compliance check, got: ${kwargs} and requires tdb, rh, tr, v, met and clo`,
+      `Missing arguments for ISO 7933:2004 compliance check, got: ${kwargs} and requires tdb, p_a, tr, v, met and clo`,
     );
   }
   /** @type {string[]} */
@@ -295,11 +298,7 @@ function _iso7933_compliance(kwargs) {
     warnings.push(
       "ISO 7933:2004 air temperature applicability limits between 15 and 50 ºC",
     );
-
-  const p_sat_result = p_sat(kwargs.tdb);
-  const p_a = ((p_sat_result / 1000) * kwargs.rh) / 100;
-
-  if (p_a > 4.5 || p_a < 0)
+  if (kwargs.p_a > 4.5 || kwargs.p_a < 0)
     warnings.push(
       "ISO 7933:2004 p_a applicability limits between 0 and 4.5 kPa",
     );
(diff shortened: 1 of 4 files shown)


src/utilities/utilities.js
```diff
@@ -71,7 +71,7 @@ export function transpose_sharp_altitude(sharp, altitude) {
  */
 
 /**
- * @typedef {"ANKLE_DRAFT" | "ASHRAE" | "FAN_HEATWAVES" | "ISO" | "ISO7933"} Standard
+ * @typedef {"ANKLE_DRAFT" | "ASHRAE" | "FAN_HEATWAVES" | "ISO" | "7933-2004" | "7933-2023"} Standard
```
FedericoTartarini: I do not like we have the same string here and in line 95, can we make use of a single class? I do not like repetitions since they make the code hard to maintain. this needs to be fixed also for the other types if possible

tests/utilities/utilities.test.js
```diff
@@ -504,55 +504,59 @@ describe("validateInputs", () => {
...
-    expect(warnings[0]).toMatch(/tr/i);
-  });
 
-  test("does not warn when tr < tdb but tr is within [0, 60]", () => {
-    // tr=20 < tdb=40, old buggy check (tr-tdb < 0) would have warned
-    const warnings = check_standard_compliance("ISO7933", {
-      ...base,
-      tr: 20,
+    test("does not warn when tr < tdb but tr is within [0, 60]", () => {
+      // tr=20 < tdb=40; old buggy check (tr-tdb < 0) would have warned
```
FedericoTartarini, line 545: make sure it is formatted with Prettier, just checking

FedericoTartarini (changes requested):
@JitaoFu I left a few minor comments

src/utilities/utilities.js
JitaoFu: Addressed in 40f4994 — extracted the standard identifiers into a frozen `Standard` constant and derived the typedef from it, so the names live in exactly one place. The switch in `check_standard_compliance` now references `Standard.ANKLE_DRAFT`, `Standard.ISO_7933_2023`, etc. instead of repeating the string literals.

tests/utilities/utilities.test.js
JitaoFu: Confirmed — `npx prettier tests/utilities/utilities.test.js --check` passes.

JitaoFu:
Looking at "this needs to be fixed also for the other types if possible" — I see three groups that could get the same treatment:

1. **Same typedef + switch pattern as `Standard`**: `clo_typical_ensembles` (9 ensemble names), `body_surface_area` (4 BSA formulas).
2. **Existing `Standard` callers still passing bare strings**: `ankle_draft.js`, `set_tmp.js`, `use_fans_heatwave.js`, `vertical_tmp_grad_ppd.js`, plus 13+ occurrences inside `phs.js`.
3. **`clo_dynamic`'s `"ASHRAE" | "ISO"`** — can reuse the new `Standard` directly instead of introducing a separate constant.

(Skipping `"IP" | "SI"` in `units_converter` — only two well-known industry abbreviations, low value-to-noise.)

Since this PR is scoped as `fix(phs): ...`, I'd lean towards a follow-up refactor PR to keep the bug fix focused, but happy to bundle them here if you prefer.

FedericoTartarini (approved):

PR #174, opened 2026-05-12
Add wind_chill_temperature model

TIAN-TOM:
Adds `wind_chill_temperature(tdb, v)`, which returns the Wind Chill Temperature in °C using the North American formula adopted by the US National Weather Service and Environment Canada. Mirrors `pythermalcomfort` 3.9.3 `wind_chill_temperature`. Motivated by comfort-tool, which currently computes WCT manually because jsthermalcomfort only exposes `wc()` (Wind Chill Index in W/m², ASHRAE / Siple-Passel), a different quantity.

The signature is `wind_chill_temperature(tdb, v, { round_output = true })`, returning `{ wct }`. `tdb` is in °C and `v` is in km/h, matching `pythermalcomfort`; this differs from the existing `wc()` function, which uses m/s. `round_output` defaults to `true`, rounding to one decimal place.

The NWS standard applicability is `tdb ≤ 10 °C` and `v ≥ 4.8 km/h`. Following `pythermalcomfort`, this implementation does not gate inputs; the formula is evaluated for any finite input.

Shared validation data has been proposed in FedericoTartarini/validation-data-comfort-models#10. Until that is merged this PR keeps the WCT reference rows local to avoid depending on an unmerged fixture branch in CI. The Jest suite also includes a tight binary check (`|wct - -7.527137645688018| < 1e-10`) on the unrounded path since the row tolerance (0.1) is too loose to lock that path alone.

### DIFF
--- a/src/models/wind_chill_temperature.js
+++ b/src/models/wind_chill_temperature.js
@@ -19,10 +19,7 @@ import { round, validateInputs } from "../utilities/utilities.js";
  *
  * @param {number} tdb - dry-bulb air temperature, [°C]
  * @param {number} v - wind speed 10 m above ground level, [km/h]
- * @param {object} [kwargs] (Optional) Other parameters.
- * @param {boolean} [kwargs.round_output=true] - if true, rounds the
- *   returned value to one decimal place; if false, returns the unrounded
- *   value.
+ * @param {boolean} [round_output=true] - if true, rounds the returned value to one decimal place; if false, returns the unrounded value.
  * @returns {{wct: number}} wind chill temperature, [°C]
  */
 const WCT_SCHEMA = {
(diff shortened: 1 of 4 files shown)


TIAN-TOM:
@FedericoTartarini This adds `wind_chill_temperature()` returning `{ wct }` for comfort-tool. The shared validation rows are proposed in FedericoTartarini/validation-data-comfort-models#10; this PR keeps them local so CI does not depend on that fixture PR merging first.

src/models/wind_chill_temperature.js
```diff
@@ -0,0 +1,49 @@
...
+ *   value.
+ * @returns {{wct: number}} wind chill temperature, [°C]
+ */
+const WCT_SCHEMA = {
+  tdb: { type: "number" },
+  v: { type: "number" },
+  round_output: { type: "boolean", required: false },
+};
+
+export function wind_chill_temperature(tdb, v, kwargs = {}) {
```
FedericoTartarini: do we really need to have kwargs here? can't we simply have three inputs td, v, and round_output?

FedericoTartarini (changes requested):
@TIAN-TOM great work, my only comment is about the kwargs

TIAN-TOM:
@FedericoTartarini Removed the kwargs wrapper in 57d5dde. The signature is now `wind_chill_temperature(tdb, v, round_output = true)`, matching `pythermalcomfort`.

FedericoTartarini (approved):

PR #176, opened 2026-05-17
Add optional round_output to adaptive_ashrae and adaptive_en

TIAN-TOM:
Adds an optional `round_output` parameter (positional, default `true`) to `adaptive_ashrae` and `adaptive_en`. With `round_output: false`, comfort temperatures and bounds are returned at full precision. Default behaviour is unchanged.

For `adaptive_ashrae`, because `t_cmf` is rounded before bounds are derived in current `main`, the unrounded path also propagates into bound derivation and can shift `acceptability_80` / `acceptability_90` at very narrow boundaries (within roughly 0.05 °C). This is intentional. `adaptive_en` doesn't have this propagation; its `t_cmf` is only rounded at the return site.

PR #166 explored a different rounding approach via cooling-effect changes and was closed; this PR keeps the cooling-effect logic unchanged. Paired upstream issue at pythermalcomfort/pythermalcomfort#323. New tests cover unrounded values, default preservation, the cooling-effect boundary, and the non-boolean type check.


TIAN-TOM:
@FedericoTartarini This adds an optional `round_output` parameter to `adaptive_ashrae` and `adaptive_en` for comfort-tool. The change is jsthermalcomfort-specific; the corresponding upstream proposal is at pythermalcomfort/pythermalcomfort#323.

FedericoTartarini (approved):

